PR #15, opened 2018-01-20
Correct wrapper library in table

nstrelow:
Just saw that typo and couldn't resist 😅 

Awesome library btw, I will definitly try it out!


derhuerst:
Thanks, this was definitely confusing for beginners!

The build is failing because of something else.

Issue #3, opened 2017-12-18
1.16 HAFAS protocol: find out what the mic & mac query parameters are for


derhuerst:
Version `1.16` of the protocol, which let's us get a lot more information from the server, seems to use two mandatory query parameters called `mic` and `mac`.

I've recorded a few requests: https://gist.github.com/derhuerst/bf3caf8685b69c44f16f073402a59e32

Hypothesis: They are hashes of the request body, salted. Similar to [how DB does it](https://github.com/derhuerst/hafas-client/blob/9f0db98ceab2ad1c33a7722cb5c2d8070220b482/p/db/index.js#L26-L35).

derhuerst:
cadd4105d11939ffe57031b4d7ed97d58d288900, a fix for derhuerst/vbb-rest#24, changed the endpoint to the BVG one. We will have to change this back.

derhuerst:
Things to be done that are available with the new protocol:

- [ ] `locL[].meta` & `locL[].isMainMast` & `locL[].mMastLocX`
	- https://gist.github.com/derhuerst/6ba0fdd6dc0d940c594f28cde067ad22#file-02-semi-accessible-http-L148
	- https://gist.github.com/derhuerst/6ba0fdd6dc0d940c594f28cde067ad22#file-02-semi-accessible-http-L150
	- https://gist.github.com/derhuerst/6ba0fdd6dc0d940c594f28cde067ad22#file-02-semi-accessible-http-L166
- [ ] "Haltestellen in der Umgebung berücksichtigen"
	- https://gist.github.com/derhuerst/6ba0fdd6dc0d940c594f28cde067ad22#file-01-take-close-stations-into-account-http-L47
- [ ] `prodL[].pid`
	- https://gist.github.com/derhuerst/6ba0fdd6dc0d940c594f28cde067ad22#file-01-take-close-stations-into-account-http-L116
- [ ] `prodL[].prodCtx.matchId`
	- https://gist.github.com/derhuerst/6ba0fdd6dc0d940c594f28cde067ad22#file-01-take-close-stations-into-account-http-L127
- [ ] `prodL[].prodCtx.admin`
	- https://gist.github.com/derhuerst/6ba0fdd6dc0d940c594f28cde067ad22#file-01-take-close-stations-into-account-http-L133
- [ ] `prodL[].stat`
[...]

PR #23, opened 2018-03-04
use 1.16 HAFAS protocol

derhuerst:
PR for #3

### DIFF
--- a/docs/journeys.md
+++ b/docs/journeys.md
@@ -41,6 +41,8 @@ With `opt`, you can override the default options, which look like this:
 ```js
 {
 	when: new Date(),
+	earlierThan: null, // ref to get journeys earlier than the last query
+	laterThan: null, // ref to get journeys later than the last query
 	results: 5, // how many journeys?
 	via: null, // let journeys pass this station
 	passedStations: false, // return stations on the way?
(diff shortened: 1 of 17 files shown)


derhuerst:
The VBB endpoint does not support the `numF` parameter in combination with the new `1.16` protocol, it just fails with `CGI_READ_FAILED`. The `numF` parameter determines the number of `journey`s in the `journeys()` method.

[There is a mechanism to get subsequent journeys](https://gist.github.com/derhuerst/5fa86ed5aec63645e5ae37e23e555886#file-2-http-L34) by providing a certain key, but I haven't figured out how to use it yet. **This means that, currently, we have no proper way to ensure a certain number of `journey`s. Technically, it would be a [breaking change](https://semver.org/spec/v2.0.0.html#semantic-versioning-specification-semver) to remove the `results` option** and just return the number of `journey`s that they provide, because people might expect `hafas-client` to return a certain number of journeys.

@deg0nz @jfilter @juliuste @blydro **What do you think should be done?** Release the (technically breaking change as `3.0.0`, or just as `2.3.4`.

derhuerst:
[I have figured it out.](https://github.com/derhuerst/hafas-client/issues/24#issuecomment-370248951) In the VBB case, **I will let `hafas-client` send more requests to collect the number of results.** Although this may lead to e.g. 10 requests to fulfill a `journeys()` call, I can circumvent a breaking change.

Sorry for pinging you all.

PR #25, opened 2018-03-04
query subsequent/preceeding journeys by ref

### LINKED ISSUE #24: journeys: find out how the subsequent/preceeding journeys feature works
[There is a mechanism to get subsequent/preceeding journeys](https://gist.github.com/derhuerst/5fa86ed5aec63645e5ae37e23e555886#file-2-http-L34), given a string that identifies/describes the last request.

Find out how this works, in order to solve https://github.com/derhuerst/hafas-client/pull/23#issuecomment-370246163 .

derhuerst:
closes #24

### DIFF
--- a/docs/journeys.md
+++ b/docs/journeys.md
@@ -249,7 +249,7 @@ Some [profiles](../p) are able to parse the ticket information, if returned by t
 
 If a journey leg has been cancelled, a `cancelled: true` will be added. Also, `departure`/`departureDelay`/`departurePlatform` and `arrival`/`arrivalDelay`/`arrivalPlatform` will be `null`.
 
-To get more journeys earlier/later than the current set of results, use `journey.earlierRef`/`journey.laterRef` as follows:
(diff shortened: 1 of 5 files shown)


derhuerst:
The build currently fails due to the notoriously unreliable "stopover at" test.

derhuerst:
todo: rename ref field to `earlierThan`/`laterThan`?

docs/journeys.md
```diff
@@ -242,3 +248,31 @@ Some [profiles](../p) are able to parse the ticket information, if returned by t
 ```
 
 If a journey leg has been cancelled, a `cancelled: true` will be added. Also, `departure`/`departureDelay`/`departurePlatform` and `arrival`/`arrivalDelay`/`arrivalPlatform` will be `null`.
+
+To get more journeys earlier/later than the current set of results, use `journey.earlierRef`/`journey.laterRef` as follows:
+
+```js
```
derhuerst: @juliuste Do you think the example code is intuitive?

derhuerst:
@kiliankoe Do you think this is an intuitive API? What would you do differently?

PR #32, opened 2018-03-16
new products markup

derhuerst:
This reduces the amount of duplicated code a lot.

- There's a single array of products per profile now. No more `products.allProducts`, `products.bitmasks` and `defaultProducts` anymore.
- The `line.mode` and `line.product` parsing has been moved into the main code. Profiles now only need to override `parseLine` if they want additional fields (or change parsed values).

**This is a breaking change** and therefore to be merged into `next` instead of `master`.


format/products-filter.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+const isObj = o => o !== null && 'object' === typeof o && !Array.isArray(o)
+const hasProp = (o, k) => Object.prototype.hasOwnProperty.call(o, k)
```
juliuste: Optional: One could use `lodash.isobject` and `lodash.has` here.

index.js
```diff
@@ -3,15 +3,19 @@
...
+const validateProfile = require('./lib/validate-profile')
 const _request = require('./lib/request')
 
 const isObj = o => o !== null && 'object' === typeof o && !Array.isArray(o)
 const isNonEmptyString = str => 'string' === typeof str && str.length > 0
 
 const createClient = (profile, request = _request) => {
 	profile = Object.assign({}, defaultProfile, profile)
+	profile.parseProducts = createParseBitmask(profile)
+	profile.formatProductsFilter = createFormatProductsFilter(profile)
```
juliuste: Wouldn't this make me unable to overwrite those two functions with my own profile? Is this intended?

derhuerst: you're right. will change it.

index.js
```diff
@@ -13,9 +13,10 @@ const isObj = o => o !== null && 'object' === typeof o && !Array.isArray(o)
 const isNonEmptyString = str => 'string' === typeof str && str.length > 0
 
 const createClient = (profile, request = _request) => {
-	profile = Object.assign({}, defaultProfile, profile)
-	profile.parseProducts = createParseBitmask(profile)
-	profile.formatProductsFilter = createFormatProductsFilter(profile)
+	profile = Object.assign({
+		parseProducts: createParseBitmask(profile),
+		formatProductsFilter: createFormatProductsFilter(profile)
+	}, defaultProfile, profile)
```
juliuste: Yes, but now the default `parseProduct` won't use the submitted `profile`'s custom products, right? 😄

derhuerst: also correct. 😛

juliuste: Ah, I see 😄

PR #33, opened 2018-03-16
radar(north, west, south, east) -> radar({north, west, south, east})

### LINKED ISSUE #27: change radar signature to radar({north, east, south, west}, opt)
to avoid confusion. Or maybe `{minLat, maxLat, minLon, maxLon}`.

derhuerst:
**This is a breaking change** and therefore to be merged into `next` instead of `master`. Closes #27.

### DIFF
--- a/index.js
+++ b/index.js
@@ -289,7 +289,7 @@ const createClient = (profile, request = _request) => {
 		})
 	}
 
-	const radar = ({north, west, south, east, opt}) => {
+	const radar = ({north, west, south, east}, opt) => {
 		if ('number' !== typeof north) throw new Error('north must be a number.')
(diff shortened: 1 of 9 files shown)


derhuerst:
build fails because of something else.

index.js
```diff
@@ -289,7 +289,7 @@ const createClient = (profile, request = _request) => {
 		})
 	}
 
-	const radar = (north, west, south, east, opt) => {
+	const radar = ({north, west, south, east, opt}) => {
```
juliuste: This should be `({north, west, south, east}, opt) => {}` instead, right?

derhuerst: yup. will fix it.

Issue #29, opened 2018-03-13
refactor products/modes markup


derhuerst:
The current markup is a mess, spread over [two](https://github.com/derhuerst/hafas-client/blob/828ef4ad47df31f303ad898d376717252126e076/p/db/modes.js) [places](https://github.com/derhuerst/hafas-client/blob/828ef4ad47df31f303ad898d376717252126e076/p/db/index.js#L106-L117) and with lots of explicitly or implicitly redundant values.

Let's make a new one. It might look like this:

```
// products.js
module.exports = [
  {
    id: 'subway',
    mode: 'train',
    name: 'U-Bahn',
    short: 'U',
    bitmasks: [64, 128], // some endpoints assign several digits to one product
    default: true
  }
  // ...
]
```

derhuerst:
This would be a breaking change because you use `hafas-client` with other profiles than the bundled ones.

derhuerst:
The `unknown` product should also be moved from the markup into the parsing code.

juliuste:
By the way, we should agree on what the `name` key should represent. There's always the "original" names (mostly german), e.g. "Durchgangszug", which often don't have an accurate translation in English.

derhuerst:
Looking at it theoretically, I think it would be good to have multiple names (at least the local one and English) for a product. But maybe this would be a bit over the top.

In the ["Writing a profile" guide](https://github.com/derhuerst/hafas-client/blob/9b6c07e477f736660c7f7cc53a59e869220eb551/docs/writing-a-profile.md#3-products), I've used the following phrase:

> A short, but distinct name for the means of transport, just precise enough in local context. In Berlin, S-Bahn commuter rail would be too much, because everyone knows what S-Bahn means.

Maybe we should add that it is about the name in the local language/context.

Issue #21, opened 2018-02-26
create simple boilerplate profile


jfilter:
to help others get started.

derhuerst:
depends on #29

derhuerst:
will do this in `next`.

PR #37, opened 2018-03-17
add a "writing a profile" guide

derhuerst:
see also #21

The products todos are left by intention, as #29 will change the products structure with `hafas-client@next` anyways. Please review everything else.


derhuerst:
The tests are not failing because of this, the PR only changes documentation.

docs/readme.md
```diff
@@ -0,0 +1,13 @@
+# API documentation
+
+- [`journeys(from, to, [opt])`](journeys.md) – get journeys between locations
```
juliuste: this should be `journeys(origin, destination, [opt])` according to FPTF

docs/writing-a-profile.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Per endpoint, `hafas-client` has an endpoint-specific customisation called *profile*** which may for example do the following:
+
+- handle the additional requirements of the endpoint (e.g. authentication),
+- extract additional information from the data provided by the endpoint,
+- guard against triggering bugs of certain endpoints (e.g. time limits).
+
+This guide is about writing such a profile. If you just want to use an already supported endpoint, refer to the [API documentation](readme.md) instead.
+
+*Note*: **If you get stuck, ask for help by [creating an issue](https://github.com/derhuerst/hafas-client/issues/new)!** I want to help people expand the scope of this library.
```
juliuste: Optional: Change `I want to` to `We want to`, if you want.

docs/writing-a-profile.md
```diff
@@ -0,0 +1,166 @@
...
+
+```js
+const myProfile = {
+	endpoint: 'https://example.org/bin/mgate.exe',
+	locale: 'de-AT',
+	timezone: 'Europe/Vienna'
+}
+```
+
+If you pass this profile into `hafas-client`, the `parseLine` method will override [the default one](../parse/line.js).
```
juliuste: > If you pass this profile into `hafas-client`…

This is a bit confusing since `myProfile` doesn't contain any `parseLine` method (yet)

docs/writing-a-profile.md
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return parseLineWithoutFoo
+}
+
+profile.parseLine = createParseLineWithoutFoo
+```
+
+## 1. Setup
+
+*Note*: There are many ways to find the required values. This way is rather easy and has worked for most of the apps that I've looked at so far.
```
juliuste: Optional: `I've looked at` -> `We've looked at`

docs/writing-a-profile.md
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+```
+
+Let's break this down:
+
+- `product`: A sensible, [camelCased](https://en.wikipedia.org/wiki/Camel_case#Variations_and_synonyms), alphanumeric identifier. Use it for the key in the `products` object as well.
+- `mode`: A [valid *Friendly Public Transport Format* `1.0.1` mode](https://github.com/public-transport/friendly-public-transport-format/blob/1.0.1/spec/readme.md#modes).
+- `bitmask`: HAFAS endpoints work with a [bitmask](https://en.wikipedia.org/wiki/Mask_(computing)#Arguments_to_functions) that toggles the individual products. the value should toggle the appropriate bit(s) in the bitmask (see below).
+- `name`: A short, but distinct name for the means of transport, *just precise enough in local context*. In Berlin, `S-Bahn commuter rail` would be too much, because everyone knows what `S-Bahn` means.
```
juliuste: If this - as discussed before - should be in the local language, we should mention this here.

docs/writing-a-profile.md
```diff
@@ -0,0 +1,166 @@
...
+As shown in [the video](https://stuff.jannisr.de/how-to-record-hafas-requests.mp4), search for a journey and toggle off one product at a time, recording the requests. After extracting the products bitmask ([example](https://gist.github.com/derhuerst/193ef489f8aa50c2343f8bf1f2a22069#file-via-http-L34)) you will end up with values looking like these:
+
+```
+toggles                     value  binary  subtraction     bit(s)
+all products                31     11111   31 - 0
+all but ACME Commuter Rail  15     01111   31 - 2^4        2^4
+all but Foo Bar Metro       23     10111   31 - 2^3        2^3
+all but product E           30     11001   31 - 2^2 - 2^1  2^2, 2^1
+all but product F           253    11110   31 - 2^1        2^0
+```
```
juliuste: It might actually be even easier to understand if you don't do 'all but product x' but rather 'only product x' since that directly gives you the corresponding powers of 2. Especially easier to understand if one product has multiple values (like for ÖBB).

docs/readme.md
derhuerst: This PR just moved it from `/readme.md` into `/docs/readme.md`. Will do the change on `master`.

docs/writing-a-profile.md
derhuerst: Ah. I've written the guide in the "I" form for now, but I'm happy to change to "we".

docs/readme.md
juliuste: See `docs/journeys.md` as well.

docs/writing-a-profile.md
derhuerst: Then it would be misaligned with the tutorial video though. Do you think it would still be more intuitive? Is it hard to understand right now?

juliuste: Ah, I didn't watch the video. Keep it like this then, it's fine enough.

docs/writing-a-profile.md
```diff
@@ -111,12 +111,12 @@ Let's break this down:
 - `product`: A sensible, [camelCased](https://en.wikipedia.org/wiki/Camel_case#Variations_and_synonyms), alphanumeric identifier. Use it for the key in the `products` object as well.
 - `mode`: A [valid *Friendly Public Transport Format* `1.0.1` mode](https://github.com/public-transport/friendly-public-transport-format/blob/1.0.1/spec/readme.md#modes).
 - `bitmask`: HAFAS endpoints work with a [bitmask](https://en.wikipedia.org/wiki/Mask_(computing)#Arguments_to_functions) that toggles the individual products. the value should toggle the appropriate bit(s) in the bitmask (see below).
-- `name`: A short, but distinct name for the means of transport, *just precise enough in local context*. In Berlin, `S-Bahn commuter rail` would be too much, because everyone knows what `S-Bahn` means.
+- `name`: A short, but distinct name for the means of transport, *just precise enough in local context*, and in the local language. In Berlin, `S-Bahn commuter rail` would be too much, because everyone knows what `S-Bahn` means.
```
juliuste: `S-Bahn commuter rail` should be translated to german then, though 😄 `Stadtschnellbahn`, for example.

derhuerst: Makes sense; Let's just pick `S-Bahn` then.

docs/writing-a-profile.md
```diff
@@ -111,7 +111,7 @@ Let's break this down:
 - `product`: A sensible, [camelCased](https://en.wikipedia.org/wiki/Camel_case#Variations_and_synonyms), alphanumeric identifier. Use it for the key in the `products` object as well.
 - `mode`: A [valid *Friendly Public Transport Format* `1.0.1` mode](https://github.com/public-transport/friendly-public-transport-format/blob/1.0.1/spec/readme.md#modes).
 - `bitmask`: HAFAS endpoints work with a [bitmask](https://en.wikipedia.org/wiki/Mask_(computing)#Arguments_to_functions) that toggles the individual products. the value should toggle the appropriate bit(s) in the bitmask (see below).
-- `name`: A short, but distinct name for the means of transport, *just precise enough in local context*, and in the local language. In Berlin, `S-Bahn commuter rail` would be too much, because everyone knows what `S-Bahn` means.
+- `name`: A short, but distinct name for the means of transport, *just precise enough in local context*, and in the local language. In Berlin, `S-Bahn-Schnellzug` would be too much, because everyone knows what `S-Bahn` means.
```
juliuste: lol 👍 😄

PR #39, opened 2018-03-21
remove journey shorthands

### LINKED ISSUE #36: remove journey.departure, journey.origin, etc
As discussed with @juliuste this is redundant information not specified in FPTF and therefore reduces the compatibility with other FPTF libraries.

**This is a breaking change.**

derhuerst:
closes #36 

This PR also fixes the tests on `next`. 😬 (Sorry for not properly creating.)

### DIFF
--- a/test/nahsh.js
+++ b/test/nahsh.js
@@ -12,7 +12,7 @@ const validateLineWithoutMode = require('./validate-line-without-mode')
 const co = require('./co')
 const createClient = require('..')
 const nahshProfile = require('../p/nahsh')
-const {allProducts} = require('../p/nahsh/products')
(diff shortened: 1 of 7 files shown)


test/nahsh.js
```diff
@@ -12,7 +12,7 @@ const validateLineWithoutMode = require('./validate-line-without-mode')
 const co = require('./co')
 const createClient = require('..')
 const nahshProfile = require('../p/nahsh')
-const {allProducts} = require('../p/nahsh/products')
+const allProducts = require('../p/nahsh/products')
```
derhuerst: This is the fix that belongs into `next` instead of this PR.

PR #44, opened 2018-04-17
Changed Journeys to Departures.

julianbei:
the endpoint which is discussed here is departures not journeys. 
If you use journeys with this parameters, you get mad with debugging.

### DIFF
--- a/docs/departures.md
+++ b/docs/departures.md
@@ -44,7 +44,7 @@ const vbbProfile = require('hafas-client/p/vbb')
 const client = createClient(vbbProfile)
 
 // S Charlottenburg
-client.journeys('900000024101', {duration: 3})
+client.departures('900000024101', {duration: 3})
 .then(console.log)
 .catch(console.error)
 ```


derhuerst:
Thanks!

I agree that the stack trace is indeed very unhelpful:

```
/Users/j/web/hafas-client/format/location.js:9
		throw new Error('invalid location type: ' + l.type)
		^

Error: invalid location type: undefined
    at Object.formatLocation (/Users/j/web/hafas-client/format/location.js:9:9)
    at Object.journeys (/Users/j/web/hafas-client/index.js:61:16)
    at Object.<anonymous> (/Users/j/web/hafas-client/p/vbb/example.js:11:8)
    at Module._compile (module.js:649:30)
    at Object.Module._extensions..js (module.js:660:10)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:501:12)
    at Function.Module._load (module.js:493:3)
    at Function.Module.runMain (module.js:690:10)
    at startup (bootstrap_node.js:194:16)
```

PR #46, opened 2018-04-23
Nah.sh: enable radar

derhuerst:
despite the failing and skipped test.

We can make this specific test handle missing station names and products, if we want. Or we fix this along with the tests rewrite.


derhuerst:
oops, compared with `master`.

Issue #34, opened 2018-03-17
nah.sh: radar() randomly swallows station names


derhuerst:
```js
const createClient = require('hafas-client')
const nahshProfile = require('hafas-client/p/nahsh')
const {inspect} = require('util')

const client = createClient(nahshProfile)

client.radar(54.4, 10.0, 54.2, 10.2)
.then((data) => {
	console.log(inspect(data, {depth: null}))
})
.catch(console.error)
```

```js
{
	direction: 'Kiel Bahnhofsvorplatz',
	trip: 31655,
	line: { /* … */ },
	location: { /* … */ },
	nextStops: [ {
		station: {
			type: 'station',
			id: '3490019',
			name: 'Hamburg Airport',
			location: { /* … */ }
		},
		departure: '2018-03-17T14:15:00.000+01:00',
		departureDelay: null,
		arrival: null,
		arrivalDelay: null
	}, {
		station: {
			type: 'station',
			id: '3435009',
			name: 'Neumünster Holstenhallen',
			location: { /* … */ }
		},
		departure: '2018-03-17T15:10:00.000+01:00',
		departureDelay: null,
		arrival: '2018-03-17T15:10:00.000+01:00',
		arrivalDelay: null
	}, {
		station: {
			type: 'station',
			id: '3440249',
			name: 'Kiel Bahnhofsvorplatz',
			location: { /* … */ }
		},
		departure: null,
		departureDelay: null,
		arrival: '2018-03-17T16:06:00.000+01:00',
		arrivalDelay: null
	}, {
		station: {
			type: 'station',
			id: '3440249',
			name: 'Kiel Bahnhofsvorplatz',
			location: { /* … */ }
		},
		departure: null,
		departureDelay: null,
		arrival: '2018-03-17T16:06:00.000+01:00',
		arrivalDelay: null
	} ],
	frames: [ {
		origin: {
			type: 'station',
			id: '3491419',
			name: undefined, // <----
			location: {
				type: 'location',
				latitude: 54.225557,
				longitude: 10.012491
			}
		},
		destination: {
			type: 'station',
			id: '3491418',
			name: undefined, // <----
			location: {
				type: 'location',
				latitude: 54.310173,
				longitude: 10.066642
			}
		},
		t: 0
	}, {
		origin: {
			type: 'station',
			id: '3491419',
			name: undefined, // <----
			location: {
				type: 'location',
				latitude: 54.225557,
				longitude: 10.012491
			}
		},
		destination: {
[...]

PR #47, opened 2018-04-23
Nah.sh: enable radar

derhuerst:
despite the failing and skipped test (see also #34).

We can make this specific test handle missing station names and products, if we want. Or we fix this along with the tests rewrite.


juliuste:
We should un-`tape.skip` the radar test as well. Could you do that?

derhuerst:
I've fixed it now. What a mess!

juliuste:
Can this be merged?

derhuerst:
if you're happy with the code, yes

juliuste:
👍

PR #50, opened 2018-05-16
expose journeyLeg/radar polyline as GeoJSON

derhuerst:
with annotated station info in each `Point`'s `properties`.

build failing because of something else


derhuerst:
ping @juliuste

PR #54, opened 2018-05-28
opt.when -> opt.departure/opt.arrival

derhuerst:
Allows one to query journeys *before* a certain timestamp.

This is a breaking change, because I removed `opt.when` completely for the sake of simplicity. It will now throw if `opt.when !== undefined`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,8 +69,8 @@ const createClient = (profile, request = _request) => {
 			if (!isNonEmptyString(opt.earlierThan)) {
 				throw new Error('opt.earlierThan must be a non-empty string.')
 			}
-			if ('when' in opt) {
-				throw new Error('opt.earlierThan and opt.when are mutually exclusive.')
+			if (('departure' in opt) || ('arrival' in opt)) {
+				throw new Error('opt.earlierThan and opt.departure/opt.arrival are mutually exclusive.')
 			}
 			journeysRef = opt.earlierThan
 		}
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -69,17 +69,17 @@ const createClient = (profile, request = _request) => {
 			if (!isNonEmptyString(opt.earlierThan)) {
 				throw new Error('opt.earlierThan must be a non-empty string.')
 			}
-			if ('when' in opt) {
-				throw new Error('opt.earlierThan and opt.when are mutually exclusive.')
+			if (('departure' in opt) || ('arrival' in opt)) {
```
juliuste: shouldn't this be `&&` instead of `||`?

index.js
```diff
@@ -69,17 +69,17 @@ const createClient = (profile, request = _request) => {
...
 			}
 			journeysRef = opt.earlierThan
 		}
 		if ('laterThan' in opt) {
 			if (!isNonEmptyString(opt.laterThan)) {
 				throw new Error('opt.laterThan must be a non-empty string.')
 			}
-			if ('when' in opt) {
-				throw new Error('opt.laterThan and opt.when are mutually exclusive.')
+			if (('departure' in opt) || ('arrival' in opt)) {
```
juliuste: same here

index.js
derhuerst: We want to check if `earlierThan`/`laterThan` and `departure`/`arrival` are being used together. The code is somewhat inelegant but – afaict – does what is supposed to.

juliuste: Sorry, I didn't read correctly. What I mistook this for is: Shouldn't `departure` and `arrival` be mutually exclusive as well?

PR #57, opened 2018-05-31
rewrite the tests

derhuerst:
Our tests currently have (at least) three problems:

1. We have to test some behaviours for every single endpoint/profile. There's a lot of duplicated code and the tests for each profile differ in how much they test.
2. The tests often only "scratch the surface" of the data returned by `hafas-client`. They often don't validate nested objects.
3. Checking a live public transport endpoint for reasonable results is inherently unstable.

**This PR attempts to address 1 & 2**.

- It introduces `validate-fptf@2`, which let's you override individual validators (e.g. the one for `location`s) so that your additional/changed validation logic will be applied to nested items as well.
- It adds helpers such as `test/lib/journeys-station-to-address.js`, which perform a single test. They reduce code duplicated across `test/*.js`.

---

A summary of the commits that will hopefully make it easier to review:

1. refactoring, preparations – 41b574c, 486929a
2. add new validation logic, add more validators, `validate-fptf@2` – 08a3349, a2abf3b, a92afc3, 2ffcf02
3. use the new validators – 1fb6d64, 46ce4bb, 0ce5e91, e1f7e07, 279dfa4
4. use the new test helpers from `test/lib` – 63e303b, 6818635, 1018369, 9501dab, cb5e01e, fecd2ae, c769638, 7f02bfe, 
2ca7e64
5. adapt the tests to be less flaky – a171234

The rest of the commits come from merging/cherry-picking from `next`.


derhuerst:
Although, in total, this PR removes only ~300 LOC, it also makes the validation a lot more thorough.

juliuste:
Are you still working on this, or should I review?

derhuerst:
please review.

PR #63, opened 2018-06-21
refactor parse functions to `parse(profile, opt, data)`

derhuerst:
This PR aligns the signatures of all higher-level parsers.

- Passing in `profile` and always calling `profile.parse…()` allows people to override single parsers (e.g. `parseLocation`) so that even deeply nested locations will be parsed with the custom logic.
- Adding `opt` allows us to add options like `parsePolyline` or `parseStationLines` to `hafas-client` in the future without changing the signature again.


derhuerst:
I didn't change every parse fn though. For the low-level ones it felt like overengineering.

Issue #56, opened 2018-05-28
"PARSE" Error when using with react-native


guitar9:
Hi i get following error when i use it with react native. This is only with client.journys(..).  client.locations(..)  works fine.

This error is only when i am not in dubugging mode. When i am in debugging mode it works.
This is my code:

```
const createClient = require('hafas-client')
const dbProfile = require('hafas-client/p/db')
const client = createClient(dbProfile)
const münchenHbf = {
  type: 'station',
  id: '008000261',
  name: 'München Hbf',
  locations: {
    type: 'location',
    latitude:48.140364,
    longitude:11.558735
  }
}

const münchenPassing = {
  type: 'station',
  id: '008004158',
  name: 'München-Pasing',
  locations: {
    type: 'location',
    latitude:48.150036,
    longitude:11.461624
  }
}
const opt = {
  //var date2 = new Date('2018-12-17T03:24:00');
	when: new Date('2018-05-29T06:40:00'),
	whenRepresents: 'departure', // use 'arrival' for journeys arriving before `when`
	//earlierThan: null, // ref to get journeys earlier than the last query
	//laterThan: null, // ref to get journeys later than the last query
	results: 5, // how many journeys?
	via: null, // let journeys pass this station
	passedStations: false, // return stations on the way?
	transfers: 5, // maximum of 5 transfers
	transferTime: 0, // minimum time for a single transfer in minutes
	accessibility: 'none', // 'none', 'partial' or 'complete'
	bike: false, // only bike-friendly journeys
	products: {
		// these entries may vary from profile to profile
		suburban: true,
		subway: true,
		tram: true,
		bus: true,
		ferry: true,
		//express: true,
		regional: true
	},
	tickets: false, // return tickets? only available with some profiles
	polylines: false // return a shape for each leg?
}
    client.journeys(münchenPassing, münchenHbf, opt)
    .then((arr)=>{
      console.log('===============arr=====================');
      console.log(arr);
      console.log('====================================');

    })
    .catch(console.error)


  }
```

```
PARSE
request
    request.js:54:13
more
    index.js:134:18
journeys
    index.js:168:14
componentDidMount
    App.js:130:20
proxiedComponentDidMount
    createPrototypeProxy.js:61:45
commitLifeCycles
    ReactNativeRenderer-dev.js:10627:12
commitAllLifeCycles
    ReactNativeRenderer-dev.js:12512:10
invokeGuardedCallback
    ReactNativeRenderer-dev.js:39:15
[...]

PR #64, opened 2018-06-24
remove Luxon from /format/date and /format/time

guitar9:
remove Luxon from /format/date and /format/time
seems to work for me now.


derhuerst:
`luxon` is most likely not the issue. Closing becuase of https://github.com/public-transport/hafas-client/issues/56#issuecomment-399915315.

PR #66, opened 2018-06-29
rename `hafas.journeyLeg()` to `hafas.trip()`

derhuerst:
Contrary to what I initially thought, this HAFAS call returns a trip anyways.


derhuerst:
@juliuste ping

PR #69, opened 2018-07-19
introduce mandatory `User-Agent` param

derhuerst:
Partly motivated by the recent eMail by VBB and partly because I think more transparent usage of the HAFAS APIs is good, **I added a mandatory `User-Agent` parameter**. People can obviously still put random words, but I assume it nudges them to pick meaningful values.


juliuste:
ehm 😄

PR #85, opened 2018-10-08
journeys from a trip to a location/station/stop

derhuerst:
Not sure if it's a DB-specific feature or just disabled/unsupported in those apps that I checked.


derhuerst:
CI is failing right now because `journeysFromTrip` doesn't work for past or future trips (only right now), so the test depends on which trains are our there. 🙄

derhuerst:
@juliuste ping

derhuerst:
@juliuste ping

derhuerst:
marudor.de implementation: https://github.com/marudor/BahnhofsAbfahrten/blob/ce0f9d3abf5b505a242f63fdc0f359af9ae2a6b6/src/server/HAFAS/SearchOnTrip.ts

juliuste (approved):

docs/journeys-from-trip.md
```diff
@@ -0,0 +1,45 @@
+# `journeysFromTrip(tripId, previousStopover, to, [opt])`
```
juliuste, line 1: Just for my understanding: Is the `previousStopover` always ignored internally and overwritten by the last one that was passed, anyway, or does it make a difference? (I'm asking because you write below that the DB Navigator always passes the first one, but AFAIK the navigator then shows results that indicate the server picked the last-passed one anyway)

derhuerst: I don't really know. 🤷

PR #88, opened 2018-10-25
added optional walkingSpeed-parameter

plSln:
gets converted to `gisFltrL`-object.
possible values: `slow`, `normal`, `fast`
default: `normal`

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,7 +37,7 @@ const createClient = (profile, userAgent, request = _request) => {
 	validateProfile(profile)
 
 	if ('string' !== typeof userAgent) {
-		 throw new Error('userAgent must be a string');
+		throw new Error('userAgent must be a string');
 	}
 
 	const _stationBoard = (station, type, parser, opt = {}) => {
--- a/index.js
+++ b/index.js
@@ -37,7 +37,7 @@ const createClient = (profile, userAgent, request = _request) => {
 	validateProfile(profile)
 
 	if ('string' !== typeof userAgent) {
-		// throw new Error('userAgent must be a string');
+		 throw new Error('userAgent must be a string');
 	}
 
 	const _stationBoard = (station, type, parser, opt = {}) => {
--- a/index.js
+++ b/index.js
@@ -205,6 +205,7 @@ const createClient = (profile, userAgent, request = _request) => {
 				getPT: true, // todo: what is this?
 				getIV: false, // todo: walk & bike as alternatives?
 				getPolyline: !!opt.polylines
+				// todo: `getConGroups: false` what is this?
 			}
 			if (profile.journeysNumF) query.numF = opt.results
 			return request(profile, userAgent, opt, {
--- a/index.js
+++ b/index.js
@@ -174,11 +174,14 @@ const createClient = (profile, userAgent, request = _request) => {
 			throw new Error('opt.walkingSpeed must be one of these values: "slow", "normal", "fast".')
 		}
 
-		const gisFltrL = [{
-			meta: 'foot_speed_' + opt.walkingSpeed,
-			mode: 'FB',
-			type: 'M'
-		}]
+		const gisFltrL = []
+		if (profile.journeysWalkingSpeed) {
+			gisFltrL.push({
+				meta: 'foot_speed_' + opt.walkingSpeed,
+				mode: 'FB',
+				type: 'M'
+			})
+		}
 
 		// With protocol version `1.16`, the VBB endpoint *used to* fail with
 		// `CGI_READ_FAILED` if you pass `numF`, the parameter for the number
--- a/index.js
+++ b/index.js
@@ -128,7 +128,7 @@ const createClient = (profile, userAgent, request = _request) => {
 		}
 
 		opt = Object.assign({
-			results: 5, // how many journeys?
+			results: null, // number of journeys – `null` means "whatever HAFAS returns"
 			via: null, // let journeys pass this station?
 			stopovers: false, // return stations on the way?
 			transfers: 5, // maximum of 5 transfers
@@ -176,7 +176,7 @@ const createClient = (profile, userAgent, request = _request) => {
 
 		const gisFltrL = [{
 			meta: 'foot_speed_' + opt.walkingSpeed,
-			mode: 'FB', 
+			mode: 'FB',
 			type: 'M'
 		}]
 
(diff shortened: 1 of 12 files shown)


schildbach:
Thanks for finding this out! I just implemented this in PTE via https://github.com/schildbach/public-transport-enabler/commit/0dcc147a21af4db8bcb73ca8797fb968481fd7f3. Works as intended with BVG.

Does anyone have an idea what the `gisFltrL.type` parameter is about? I had to change its value from `P` to `M`, otherwise the walk speed would be ignored.

plSln:
@schildbach ah, good to know, I wondered why it had no effect..
I will try to find that out!

derhuerst:
> Does anyone have an idea what the `gisFltrL.type` parameter is about? I had to change its value from `P` to `M`, otherwise the walk speed would be ignored.

The HAFAS `mgate.exe` looks exactly like JSON created from an XML tree, which is why AFAICT `.type` corresponds to the tag name of the corresponding XML element.

I don't know what `M` and `P` stand for though.

index.js
```diff
@@ -170,6 +170,11 @@ const createClient = (profile, userAgent, request = _request) => {
 			filters.push(profile.filters.accessibility[opt.accessibility])
 		}
 
+		const gisFltrL = [{
+			meta: ['slow','normal','fast'].includes(opt.walkingSpeed) ? 'foot_speed_' + opt.walkingSpeed : 'foot_speed_normal',
```
derhuerst: Sorry, I didn't explain clearly enough. Please throw on invalid usage, as seen above.

derhuerst:
I will try to write a test for this. Do you know if all endpoints support this (even though not every app has it in the UI)?

plSln:
> I will try to write a test for this. Do you know if all endpoints support this (even though not every app has it in the UI)?

I don't know that, Sorry.

derhuerst:
It seems to be a feature of the `1.21` protocol version. Will further look into it.

derhuerst:
I've tried to use the `walkingSpeed` option with BVG, CMTA, DB, Nah.SH, S-Bahn München & VBB, but only **apparently only the BVG & VBB endpoints make use of it.** Thus, I have made it opt-in per profile.

@schildbach Can you confirm this? Check my BVG test case as an example:

https://github.com/public-transport/hafas-client/blob/306dd9453b2321a8ce58fd7e5008f53c7530f3cb/test/bvg.js#L130-L148

https://github.com/public-transport/hafas-client/blob/306dd9453b2321a8ce58fd7e5008f53c7530f3cb/test/lib/journeys-walking-speed.js#L10-L28

schildbach:
Oh, indeed. Just checked DB and SH and it would not make a difference. How annoying!

schildbach:
On the [VOR web page](https://anachb.vor.at), I found that they set these values: `{"type":"M","mode":"FB","meta":"foot_speed_fast"},{"type":"M","mode":"T","meta":"foot_speed_fast"}`
So that's one more value, but in my tests it didn't change anything. They have speed options for bike and car too though, might be interesting.

derhuerst:
> On the VOR web page, I found that they set these values [...] but in my tests it didn't change anything. They have speed options for bike and car too though, might be interesting.

Yes, noticed that too for some other endpoints supported by `hafas-client`. We should keep an eye on that in the near future!

BTW: We should join our efforts better to cover HAFAS features! You're always welcome to open issues here to let us know about things you found out, and I will try to do the same.

derhuerst:
@plSln have you [allowed me to push into your fork](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)?

plSln:
> @plSln have you [allowed me to push into your fork](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)?

Yes!

derhuerst:
```shell
git branch -a
```

```
  berlkoenig
  cfl
  dsb
  irish-rail
  journeys-from-trip
  master
  mobiliteit-lu
  mock
  moobil-plus
  next
  saarfahrplan
  strip-leading-zeros
  tz-offset
  vbn
* walking-speed
  remotes/origin/berlkoenig
  remotes/origin/cfl
  remotes/origin/dsb
  remotes/origin/irish-rail
  remotes/origin/journeys-from-trip
  remotes/origin/master
  remotes/origin/mobiliteit-lu
  remotes/origin/mock
  remotes/origin/next
  remotes/origin/strip-leading-zeros
  remotes/origin/tz-offset
  remotes/origin/vbn
  remotes/origin/walking-speed
  remotes/plSln/cfl
  remotes/plSln/dsb
  remotes/plSln/irish-rail
  remotes/plSln/journeys-from-trip
  remotes/plSln/master
  remotes/plSln/mobiliteit-lu
  remotes/plSln/mock
  remotes/plSln/next
  remotes/plSln/scheduled-days
  remotes/plSln/strip-leading-zeros
  remotes/plSln/tz-offset
  remotes/plSln/vbn
```

derhuerst:
Got help, solved it using `git push plSln HEAD:master`.

derhuerst:
@juliuste `departures()`/`arrivals()` are still failing (as intended), but please have a look at the code.

test/lib/journeys-walking-speed.js
```diff
@@ -0,0 +1,32 @@
...
+	const legWithSlowWalking = journeyWithSlowWalking.legs.find(l => l.mode === 'walking')
+	t.ok(legWithSlowWalking, 'no walking leg in journey with slow walking')
+
+	const fastDist = legWithFastWalking.distance
+	const slowDist = legWithSlowWalking.distance
+	t.ok(isRoughlyEqual(100, fastDist, slowDist), 'precondition failed')
+	const fastDur = new Date(legWithFastWalking.arrival) - new Date(legWithFastWalking.departure)
+	const slowDur = new Date(legWithSlowWalking.arrival) - new Date(legWithSlowWalking.departure)
+	// console.error(fastDur / 1000 / 60, slowDur / 1000 / 60)
+	t.notOk(isRoughlyEqual(dT, fastDur, slowDur), 'walkingSpeed not applied')
```
juliuste: `dT` is not the most intuitive naming here IMHO 😄

lib/request.js
```diff
@@ -88,7 +88,7 @@ const request = (profile, userAgent, opt, data) => {
 	.then((b) => {
 		if (DEBUG) console.error(JSON.stringify(b))
 
-		if (b.err) {
+		if (b.err && b.err !== 'OK') {
```
juliuste: Is this correct here? Just asking because as this doesn't seem directly related to the scope of the PR I want to make sure this is an intended change (since it affects all requests)? 😄

derhuerst: This is part of the switch to the `1.21` protocol, so yes, intended. From what I can tell, it doesn't break endpoints with older versions.

derhuerst:
Sorry for the back-and-forth: `1.21` fails if you pass `stbFltrEquiv`. It would be a breaking change if we drop it.

derhuerst:
I have squashed and slightly adapted the commits into abc9c2b, 7908fb3, 61fbe15, 7073234 on the `5` branch. Thanks for your contribution!

PR #89, opened 2018-10-25
add CFL profile

derhuerst:


derhuerst:
This is WIP, I need to adapt the branch to the current code base.

p/cfl/index.js
```diff
@@ -0,0 +1,33 @@
...
+		os: 'iPhone OS 9.3.5'
+	}
+	body.ver = '1.16'
+	body.auth = {aid: 'ALT2vl7LAFDFu2dz'}
+	body.lang = 'de'
+
+	return body
+}
+
+const insaProfile = {
```
juliuste: This should be called `cflProfile`

p/cfl/index.js
```diff
@@ -0,0 +1,33 @@
...
+	locale: 'de-DE',
```
juliuste: Shouldn't this be the luxemburgish locale?

p/cfl/index.js
```diff
@@ -0,0 +1,33 @@
...
+	timezone: 'Europe/Berlin',
```
juliuste: Shouldn't this be `Europe/Luxembourg`?

PR #91, opened 2018-10-27
Fixes optional price behavior

lightsprint09:
As described in fptf amount and currency should be required. If amount is not available the whole price object is nil

```
price: { // optional
	amount: 12.50, // number, required
	currency: 'EUR' // ISO 4217 code, required
}
```

### DIFF
--- a/p/db/index.js
+++ b/p/db/index.js
@@ -42,8 +42,8 @@ const createParseJourney = (profile, opt, data) => {
 
 	// todo: j.sotRating, j.conSubscr, j.isSotCon, j.showARSLink, k.sotCtxt
 	// todo: j.conSubscr, j.showARSLink, j.useableTime
-	const parseJourneyWithPrice = (j) => {
-		const res = parseJourney(j)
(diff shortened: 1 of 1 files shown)


p/db/index.js
```diff
@@ -55,20 +55,18 @@ const createParseJourney = (profile, opt, data) => {
 		// 	targetCtx: 'D',
 		// 	buttonText: 'To offer selection'
 		// } ]
-		res.price = {amount: null, hint: 'No pricing information available.'}
```
derhuerst: Please assign `null` as the default value.

lightsprint09: Added `null` as default

derhuerst:
Technically someone could rely one the behaviour that, if `journey.price` is truthy, it will contain a price. Because of this, I'm not sure if this is a breaking change.

derhuerst:
Sorry, I just noticed that I have added this behaviour in 8677a1cc on branch `5` already. Nonetheless, thanks for your contribution!

PR #94, opened 2018-11-12
add INVG

### LINKED ISSUE #93: support INVG (Ingolstadt)
`https://fpa.invg.de/bin/mgate.exe` uses HCI. Versions above 1.13 would likely need a mic/mac salt, otherwise an AUTH error is raised. The iOS app should have that, but I haven't got an iOS device. For details see https://github.com/schildbach/public-transport-enabler/commit/dc390eb5a407f491c14481ab9c87206ed09d0f27.

derhuerst:
Closes #93.

~Tests are currently failing because, for `departures()`, we get `stop` IDs without `station` IDs ([just like with ÖBB](https://github.com/public-transport/hafas-client/blob/9d969027941be04fc6af5766cd0eea50ca3daeb2/test/oebb.js#L176)).~


p/invg/example.js
```diff
@@ -0,0 +1,43 @@
+'use strict'
+
+const createClient = require('../..')
+const cmtaProfile = require('.')
+
+const client = createClient(cmtaProfile, 'hafas-client-example')
+
+const ingolstadtHbf = '8000183'
+const audiParkplatz = '84999'
```
juliuste: xD

p/invg/products.js
```diff
@@ -0,0 +1,77 @@
...
+	{
+		id: 'bus',
+		mode: 'bus',
+		bitmasks: [1],
+		name: 'Bus',
+		short: 'Bus',
+		default: true // the other `bus` has `false`
+	},
+	{
+		id: 'express-train',
```
juliuste: *[minor]* Are you sure the naming is appropriate here? If I understand the profile correctly this includes intercity and eurocity trains, which are faster than regional, but not really high speed.

p/invg/readme.md
```diff
@@ -0,0 +1,18 @@
+# INVG profile for `hafas-client`
+
+[*Ingolstädter Verkehrsgesellschaft (INVG)*](https://de.wikipedia.org/wiki/Ingolstädter_Verkehrsgesellschaft) is a public transportation provider serving [Ingolsstadt, Germany](https://en.wikipedia.org/wiki/Ingolstadt). This profile adds *INVG*-specific customizations to `hafas-client`.
```
juliuste:
```suggestion
[*Ingolstädter Verkehrsgesellschaft (INVG)*](https://de.wikipedia.org/wiki/Ingolstädter_Verkehrsgesellschaft) is a public transportation provider serving [Ingolstadt, Germany](https://en.wikipedia.org/wiki/Ingolstadt). This profile adds *INVG*-specific customizations to `hafas-client`.
```

p/invg/products.js
```diff
@@ -0,0 +1,77 @@
...
+	{
+		id: 'local-train',
+		mode: 'train',
+		bitmasks: [8],
+		name: 'Nahverkehrszug',
+		short: 'Zug',
+		default: true
+	},
+	{
+		id: 'bus-2',
```
juliuste: The naming here seems sub-optimal 😛 IMHO we should rather group both bus products if we don't exactly know the difference between them.

p/invg/products.js
derhuerst: I didn't think much about this name, I may even have copied it from another profile. "express" here rather meant "only stopping at some stations", but we can change it to something else.

p/invg/products.js
derhuerst: I think the reason why I didn't merge them back then was that one is on by default, the other is not. Will merge them nevertheless.

PR #99, opened 2018-12-28
add Saarfahrplan profile

derhuerst:


p/saarfahrplan/example.js
```diff
@@ -0,0 +1,43 @@
+'use strict'
+
+const createClient = require('../..')
+const oebbProfile = require('.')
```
juliuste: We should rename this

p/saarfahrplan/example.js
```diff
@@ -0,0 +1,43 @@
...
+
+const client = createClient(oebbProfile, 'hafas-client-example')
```
juliuste: same here

derhuerst:
Only unrelated tests are failing.

juliuste:
By the way: We might consider using something like your `tap-min` module in the `npm test` script given the output size we have reached now.

derhuerst:
I've purposefully avoided `tap-min`, because it swallows/shortens TAP output too much. Maybe there's something like [`tap-spec`](https://github.com/scottcorgan/tap-spec#readme) that doesn't print successful assertions.

derhuerst:
@ski7777 Did you see this? I think you had asked for SaarVV support a while ago.

PR #110, opened 2019-04-24
ISO date+time: suppress milliseconds if 0

derhuerst:


derhuerst:
At work, we decided to go this way because some platforms have problems with millisecond ISO dates. As this change doesn't change any semantics and doesn't break consumers (as long as they parse ISO 8601 properly), I felt it is fitting for `hafas-client`.

juliuste:
Alright, so as I said: LGTM. Feel free to merge whenever you want.

PR #112, opened 2019-05-20
DB departures/arrivals/journeys: parse load factor

### LINKED ISSUE #111: DB: Support new DB.R19.04.a features
Since April 2019, DB Navigator and bahn.de use DB.R19.04.a  as API ext.
This adds "Auslastungsinformation" (demand in English), as mentioned on [inside.bahn.de]( https://inside.bahn.de/db-navigator-funktionen/#Auslastungsanzeige) and [bahn.de](https://www.bahn.de/p/view/service/buchung/auslastungsinformation.shtml)

Using DB.R19.04.a in hafas-client returns this information as tcocL, with references from the journeys as dTrnCmpSX.tcocX.

JSON from refreshJourney: https://gist.github.com/TheZenti/12c3d8323c7a357774c1906545f31ecc#file-journeydetails-withload-json-L252
When using departures:
https://gist.github.com/TheZenti/12c3d8323c7a357774c1906545f31ecc#file-journeydetails-withload-json-L252

I've already started implementing rudimentary support for this in [my own fork](https://github.com/TheZenti/hafas-client), but it would be nice if hafas-client officially supports this ;)

I'm not sure if there are more new features, or if some got removed. I haven't taken a look at that yet.

derhuerst:
This PR updates the `ext` value in order to parse the load factor from the response. See https://www.bahn.de/p/view/service/buchung/auslastungsinformation.shtml for an explanation about the possible values.

closes #111.

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,6 +89,7 @@ const createClient = (profile, userAgent, request = _request) => {
 		.then((d) => {
 			if (!Array.isArray(d.jnyL)) return []
 			const parse = parser(profile, opt, {
+				raw: d,
 				locations: d.locations,
 				lines: d.lines,
 				hints: d.hints,
@@ -218,6 +219,7 @@ const createClient = (profile, userAgent, request = _request) => {
 				if (!Array.isArray(d.outConL)) return []
 
 				const parse = profile.parseJourney(profile, opt, {
+					raw: d,
 					locations: d.locations,
 					lines: d.lines,
 					hints: d.hints,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -89,6 +89,7 @@ const createClient = (profile, userAgent, request = _request) => {
 		.then((d) => {
 			if (!Array.isArray(d.jnyL)) return []
 			const parse = parser(profile, opt, {
+				raw: d,
```
juliuste: It seems a bit inconsistent to pass all of `d` here instead of `d.common`

derhuerst: True. My intention is to migrate all `profile.parse*` functions to just take `d` though; I didn't want to change all in this PR.

juliuste: k

PR #114, opened 2019-05-20
add NVV profile

derhuerst:
AFAIK tests are failing because we get `stop` IDs without `station` IDs.


p/nvv/products.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+module.exports = [
+	// todo: what is `256`?
+	{
+		id: 'express',
+		mode: 'train',
+		bitmasks: [1],
+		name: 'InterCityExpress',
```
juliuste: Can this also be `TGV`, for example?

derhuerst: Apparently TGV trains are not shown in the NVV app.

PR #118, opened 2019-06-03
add leg.reachable

juliuste:
This PR exposes the `isRchbl` flag on legs as `leg.reachable`.


derhuerst:
Do you know what exactly this flag expresses? If a transit leg is reachable from the previous transit leg?

juliuste:
Yes, it seems to correspond with the green/red indicators that appear sometimes in the DB app, when the "transfer" was red, the flag was false, otherwise true. However, the Navigator app seems to do perform additional logic on top of reachable/non-reachable.

derhuerst:
Published as `hafas-client@4.2.0`.

PR #123, opened 2019-07-02
add debug CLI

derhuerst:
This tool helps you quickly pull out data from HAFAS endpoints. You can use [`jq`](https://stedolan.github.io/jq/) to pretty-print or process the JSON.


derhuerst:
It doesn't work yet with e.g. `cli.js stop 890723`, because it parses `890723` into a number, but the stop ID must be a string.

Issue #125, opened 2019-08-20
add profile for DB Busradar NRW


derhuerst:
As documented in https://gist.github.com/d3d9/350b25b65f6dc4a92195cfbb12d1cc45, the DB Busradar NRW app seems to use HAFAS with the `JourneyGeoPos` call. Let's [write a profile](https://github.com/public-transport/hafas-client/blob/3c7c1c3d4efcef83fa4e699f1fe12681585f4644/docs/writing-a-profile.md) for it!

cc @d3d9

d3d9:
Here's what I have right now (should I already make a pull request or wait until some things are fixed and added?):

### index.js
see comments above and inside `products`

```js
'use strict'

// DB Busradar NRW app does not allow selecting specific modes of transport to filter results,
// so the bitmasks had to be determined by querying some stations and looking at the results..
const products = [
	{
		id: 'nationalExpress',
		mode: 'train',
		bitmasks: [1],
		name: 'InterCityExpress',
		short: 'ICE',
		default: true
	},
	{
		id: 'national',
		mode: 'train',
		bitmasks: [2],
		name: 'InterCity & EuroCity',
		short: 'IC/EC',
		default: true
	},
	// 4: not always used, could not be confirmed..
	{
		id: 'regionalExp',
		mode: 'train',
		bitmasks: [4],
		name: 'Regionalexpress',
		short: 'RE',
		default: true
	},
	{
		id: 'regional',
		mode: 'train',
		bitmasks: [8],
		name: 'Regionalzug',
		short: 'RB/RE',
		default: true
	},
	// 16: often appears together with 8 even when there are only S-Bahn trains at the station
	{
		id: 'suburban',
		mode: 'train',
		bitmasks: [16],
		name: 'S-Bahn',
		short: 'S',
		default: true
	},
	{
		id: 'bus',
		mode: 'bus',
		bitmasks: [32],
		name: 'Bus',
		short: 'Bus',
		default: true
	},
	// 64, 128, 256: could not be confirmed
	{
		id: 'ferry',
		mode: 'watercraft',
		bitmasks: [64],
		name: 'Ferry',
		short: 'F',
		default: true
	},
	{
		id: 'subway',
		mode: 'train',
		bitmasks: [128],
		name: 'U-Bahn',
		short: 'U',
		default: true
	},
	{
		id: 'tram',
		mode: 'train',
		bitmasks: [256],
		name: 'Tram',
		short: 'T',
		default: true
	},
	{
[...]

PR #126, opened 2019-08-20
Add DB Busradar NRW profile

d3d9:
See #125.

Todo:
- [ ] fix products
- [ ] tests
- [x] readme
- ..?

### DIFF
--- a/p/db-busradar-nrw/index.js
+++ b/p/db-busradar-nrw/index.js
@@ -106,7 +106,7 @@ const transformReqBody = (body) => {
 const dbbusradarnrwProfile = {
 	locale: 'de-DE',
 	timezone: 'Europe/Berlin',
-	endpoint: 'http://db-regio.hafas.de/bin/hci/mgate.exe',
+	endpoint: 'https://db-regio.hafas.de/bin/hci/mgate.exe',
 	transformReqBody,
 
 	products: products,
--- a/p/db-busradar-nrw/index.js
+++ b/p/db-busradar-nrw/index.js
@@ -1,0 +1,119 @@
+'use strict'
+
+// DB Busradar NRW app does not allow selecting specific modes of transport to filter results,
+// so the bitmasks had to be determined by querying some stations and looking at the results..
+const products = [
+	{
+		id: 'nationalExpress',
+		mode: 'train',
+		bitmasks: [1],
+		name: 'InterCityExpress',
+		short: 'ICE',
+		default: true
+	},
+	{
+		id: 'national',
+		mode: 'train',
+		bitmasks: [2],
+		name: 'InterCity & EuroCity',
+		short: 'IC/EC',
+		default: true
+	},
+	// 4: not always used, could not be confirmed..
+	{
+		id: 'regionalExp',
+		mode: 'train',
+		bitmasks: [4],
+		name: 'Regionalexpress',
+		short: 'RE',
+		default: true
+	},
+	{
+		id: 'regional',
+		mode: 'train',
+		bitmasks: [8],
+		name: 'Regionalzug',
+		short: 'RB/RE',
+		default: true
+	},
+	// 16: often appears together with 8 even when there are only S-Bahn trains at the station
+	{
+		id: 'suburban',
+		mode: 'train',
+		bitmasks: [16],
+		name: 'S-Bahn',
+		short: 'S',
+		default: true
+	},
+	{
+		id: 'bus',
+		mode: 'bus',
+		bitmasks: [32],
+		name: 'Bus',
+		short: 'Bus',
+		default: true
+	},
+	// 64, 128, 256: could not be confirmed
+	{
+		id: 'ferry',
+		mode: 'watercraft',
+		bitmasks: [64],
+		name: 'Ferry',
+		short: 'F',
+		default: true
+	},
+	{
+		id: 'subway',
+		mode: 'train',
+		bitmasks: [128],
+		name: 'U-Bahn',
+		short: 'U',
+		default: true
+	},
+	{
+		id: 'tram',
+		mode: 'train',
+		bitmasks: [256],
+		name: 'Tram',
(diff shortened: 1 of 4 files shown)


p/db-busradar-nrw/index.js
```diff
@@ -0,0 +1,119 @@
...
+	},
+	{
+		id: 'national',
+		mode: 'train',
+		bitmasks: [2],
+		name: 'InterCity & EuroCity',
+		short: 'IC/EC',
+		default: true
+	},
+	// 4: not always used, could not be confirmed..
```
derhuerst: Please check if the `4` bit is always being toggled with another one, otherwise mark this comment as resolved. With other endpoints, two bitmasks together represent one product: https://github.com/public-transport/hafas-client/blob/16f98b943eaf58a8c7d57f049f492ef1f49b2c01/p/oebb/products.js#L15

p/db-busradar-nrw/index.js
```diff
@@ -0,0 +1,119 @@
...
+	},
+	{
+		id: 'regional',
+		mode: 'train',
+		bitmasks: [8],
+		name: 'Regionalzug',
+		short: 'RB/RE',
+		default: true
+	},
+	// 16: often appears together with 8 even when there are only S-Bahn trains at the station
```
derhuerst: As above.

p/db-busradar-nrw/index.js
```diff
@@ -0,0 +1,119 @@
...
+	},
+	{
+		id: 'bus',
+		mode: 'bus',
+		bitmasks: [32],
+		name: 'Bus',
+		short: 'Bus',
+		default: true
+	},
+	// 64, 128, 256: could not be confirmed
```
derhuerst: As above.

p/db-busradar-nrw/index.js
```diff
@@ -0,0 +1,119 @@
...
+	body.ver = '1.10'
+	body.auth = {type: 'AID', aid: 'OGBAqytjHhCvr0J4'}
+	body.lang = 'eng'
+	return body
+}
+
+const dbbusradarnrwProfile = {
+	locale: 'de-DE',
+	timezone: 'Europe/Berlin',
+	endpoint: 'http://db-regio.hafas.de/bin/hci/mgate.exe',
```
derhuerst: The endpoint supports HTTPS. Can you change this to `https:`?

p/db-busradar-nrw/index.js
d3d9: I haven't found a 4 on it's own, it seems to appear often on larger stations that have other products as well. There are stations with one of the other products toggled without 4 being toggled as well.
I can't find any hints to the actual meaning of it. If it would be for RegionalExpress, most stations would have wrong data, like Schwelm where there's an 8 but not a 4, while Hagen Hbf correctly also has a 4.
4 also does not seem to be related to something like a tram or subway, as these products do not exist at Hagen Hbf.

p/db-busradar-nrw/index.js
d3d9: The appearance together with 8 may be because of 8 being used for replacement service, even when it's for S-Bahn, for example Dormagen Chempark has a 16 on it's own while the next station, Köln Worringen, has a 16 with an 8 because there's also replacement service at that station (but no usual RB or RE)

p/db-busradar-nrw/index.js
d3d9: 64 seems to be correct, for example at Emden Außenhafen.
I can't find any station with 128 and 256 right now.

p/db-busradar-nrw/index.js
derhuerst: Even if you don't *exactly* know what 16 stands for, use the closes identifier & description you can come up with. We can improve on it later.

Also, once you add it, please set `default: true`, unless you're very sure that this product is irrelevant in the results.

p/db-busradar-nrw/index.js
derhuerst: If you, as an experience local, don't notice any departures/results missing without `4`, that's fine. 😉

d3d9: It seems to only return bus departures in any case, so that wouldn't be a problem. It being returned with different values at different stations may still be interesting though -- subway and tram always(?) return false so these could be removed without any loss of information, unlike this 4..

p/db-busradar-nrw/index.js
d3d9: 16 should be ok as is, it is clearly S-Bahn. It's just the 8 which is not very obvious, it seems to be for RB+RE+replacement services

p/db-busradar-nrw/index.js
derhuerst: Don't remove anything just because the endpoint doesn't return data for it right now. Maybe the endpoint will have the data in the future.

derhuerst:
Please add an entry to the products table in the readme. I will take care of the tests.

p/db-busradar-nrw/index.js
d3d9: Ok, so should I keep "regionalExp" as is even though we don't know whether that is even correct or can it be renamed to say that it is an unknown mode?

derhuerst: - Keep all products if you know their bitmask(s), even if they're not being returned by the endpoint right now.
- Add all products you don't know as e.g. "unknown".

d3d9:
The android app recently got updated; the UI looks different and there may be some added search options (like a departure list in a specific direction, if that was not in the UI before).

It also looks like fake position data is used more often than before and the real GPS position gets updated less often..

The app also seems to ignore the proxy settings now.

derhuerst:
I finally managed to look into this. Will adapt the products and add tests.

derhuerst:
Merged and published as `hafas-client@4.8.0`. 🎉

PR #127, opened 2019-08-23
resolve index-based references before parsing

derhuerst:
This PR makes the parsing logic in `parse/*.js` simpler and more readable by resolving index-based references in `parse/common.js` already.

Consider the resolving with this simplified HAFAS response:

```js
{
	common: {
		locL: [
			{
				type: 'S',
				name: 'S+U Jannowitzbrücke (Berlin)',
				icoX: 2, // `common.icoL[2]` inserted as `icon` here
				extId: '900100004',
				crd: {x: 13418027, y: 52515503}
			}
		],
		prodL: [
			{
				name: 'U8',
				icoX: 0, // `common.icoL[0]` inserted as `icon` here
				cls: 2,
				oprX: 0 // `common.opL[0]` inserted as `operator` here
			}
		],
		opL: [
			{
				name: 'Berliner Verkehrsbetriebe',
				icoX: 1 // `common.icoL[1]` inserted as `icon` here
			}
		],
		icoL: [
			{
				res: 'prod_sub_t'
			},
			{
				res: 'BVG',
				txt: 'Berliner Verkehrsbetriebe'
			},
			{
				res: 'prod_comm_t'
			}
		]
	},
	jnyL: [
		{
			jid: '1|31015|8|86|19082019',
			date: '20190819',
			prodX: 0, // `common.prodL[0]` inserted as `line` here
			dirTxt: 'S+U Hermannstr. (Berlin)',
			stbStop: {
				locX: 0, // `common.locL[0]` inserted as `location` here
				dTimeS: '203000',
			}
		}
	]
}
```


lib/resolve-idx-refs.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+const scanner = require('object-scan')
+const toPath = require('lodash/toPath')
+const get = require('lodash/get')
+
+const findIdxRefs = (tree, selector, onRef) => {
```
juliuste: This is pretty hard to understand (same for the method below), can you add a comment that explains roughly what the different parameters mean?

lib/resolve-idx-refs.js
```diff
@@ -0,0 +1,36 @@
...
+	}
+}
+
+// For all items in `tree` matching a path selector specified
+// in `selector`:
+// - interpret the item as an index in `source`
+// - set the parent's field `prop` to `source[item]`
+const resolveIdxRefs = (tree, selector, source, prop) => {
+	findIdxRefs(tree, selector, (idx, parent) => {
+		if ('number' === typeof idx) parent[prop] = source[idx]
```
juliuste: Ok. If I understood everything correctly, this code does what it should, however I'm - to put it mildly - not the biggest fan of mutating the original object. I think it would be cleaner (and easier to test) to have a function that extracts a property from the original tree without modifying it, but that's just a style question, and I don't want to block this PR for this.

derhuerst: > I think it would be cleaner (and easier to test) to have a function that extracts a property from the original tree without modifying it [...].

I agree in theory. In practice, letting `parse/common.js` use `findIdxRefs` directly would just move boilerplate code. Do you agree?

juliuste: I agree, but this would make it much easier to reason about what happens. In `parse/common`, I kind of have the *bigger picture* and can tell what is happening, while in this function here, I don't really see the bigger picture, but I indirecty see the function changing something in it.

derhuerst: Changed it in a3f0cee.

PR #134, opened 2019-09-16
support rest.exe endpoints

derhuerst:
This PR contains the commits of #129, so merge that first.

---

still to do:

- [ ] cover all features of the API
- [ ] unify `parse` and `parse-rest`
- [ ] find a way to write a profile for `rest.exe`
- [ ] tests
- [ ] documentation


juliuste:
(Will rebase to current master)

parse-rest/journey-leg.js
```diff
@@ -57,7 +57,7 @@ const parseJourneyLeg = (ctx, l) => { // l = leg
 		res.tripId = l.JourneyDetailRef && l.JourneyDetailRef.ref || null
 
 		// todo: is it num instead of number?
-		const product = {name: l.name, number: l.number, ...l.Product}
+		const product = {name: l.name, number: l.number, ...(l.Product[0] || {})}
```
derhuerst: I think this should be done in the [simplification/`findInTree` section](https://github.com/public-transport/hafas-client/blob/e6755ffbd18a0f1d1610cca1c0e6173b6fca7d98/rest-exe.js#L110-L128).

marzieh-kazemi:
`id` is missing on the trip returned by `.trip(id)`.

juliuste:
Reminder that this code lives in `db-hafas-rest-client` after we discussed to separate the libs some time ago (for the time being)

derhuerst:
Maybe the [`bene-system.com` `query.exe` endpoint](https://github.com/schildbach/public-transport-enabler/issues/647) has the same API?

PR #136, opened 2019-09-29
createClient(): request parameter -> opt parameter

derhuerst:


retry.js
```diff
@@ -13,7 +13,9 @@ const retryDefaults = {
 const withRetrying = (createClient, retryOpts = {}) => {
 	retryOpts = Object.assign({}, retryDefaults, retryOpts)
 
-	const createRetryingClient = (profile, userAgent, request = _request) => {
+	const createRetryingClient = (profile, userAgent, opt = {}) => {
+		const {request} = Object.assign({request: _request}, opt)
```
juliuste: You could move this to a `defaults` variable to be consistent with `index.js` (or even move this into its own function, since it's used three times), but I don't mind if you don't.

throttle.js
```diff
@@ -5,9 +5,15 @@ const throttle = require('p-throttle')
 const _request = require('./lib/request')
 
 const withThrottling = (createClient, limit = 5, interval = 1000) => {
-	const createThrottledClient = (profile, userAgent, request = _request) => {
+	const createThrottledClient = (profile, userAgent, opt = {}) => {
+		const {request} = Object.assign({request: _request}, opt)
```
juliuste: same as above

retry.js
derhuerst: Labeling this object as `defaults` sounds like it represents all options that `createClient` from `index.js` knows about. All I want to do here get the `request` fn if it has been passed in.

derhuerst: I could also change the line below to `const request = 'request' in opt ? opt.request : _request`.

Issue #146, opened 2020-01-24
Improve find-in-tree.js


simlu:
- object-scan has a callback method "filterFn" that exposes the parent (no need to loop through the results again)
- object-scan has a "joined" flag, so there is no need to use lodash toPath

https://github.com/public-transport/hafas-client/blob/42b2a8a7bfd163504317279b967aa64e81996247/lib/find-in-tree.js#L14

simlu:
Something like
```js
const findInTree = (tree, selector, onResult) => {
  scanner([selector], {
    joined: false,
    filterFn: (key, value, { parents }) => {
      onResult(value, parents[0], key);
    }
  })(tree);
}
```
(untested)

derhuerst:
PR welcome!

> (untested)

The tests should tell you wether you broke it. 😉

simlu:
@derhuerst Would be great if you could set up CI, so I can do it and test right through Github 😉

simlu:
I've created the pr. How do we get Travis to run against it?

simlu:
@derhuerst Pr is up 👍 Looks like there is no eslint test (the formatting I've used should probably fail) and the e2e test also failed?

Please do take a look

PR #147, opened 2020-01-26
chore: improved performance

simlu:
closes https://github.com/public-transport/hafas-client/issues/146

### DIFF
--- a/lib/find-in-tree.js
+++ b/lib/find-in-tree.js
@@ -1,8 +1,6 @@
 'use strict'
 
 const scanner = require('object-scan')
-const toPath = require('lodash/toPath')
-const get = require('lodash/get')
 
 // For all items in `tree` matching a path selector specified
 // in `selector`, call `onResult` with the item, its parent and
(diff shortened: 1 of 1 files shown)


derhuerst:
> Looks like there is no eslint test (the formatting I've used should probably fail) [...]?

Please just follow the style specified in `.editorconfig`: use tabs for JavaScript. Other than that, I don't have strong (enough) opinions about the style to add ESLint, your code style is fine to me.

> Looks like [...] the e2e test also failed?

Yes, that is the currently the case, unfortunately. You didn't break it though, and the "unit" tests (using fixtures) are passing. Testing against actual live transit systems in a reliable way is really hard.

simlu:
@derhuerst Sounds good. I did this pr on my phone, so didn't have tabs available...

Regarding e2e tests: Please take a look at node-tdd. It uses nock under the hook and should be exactly what you want for e2e testing (rerecord when necessary, but otherwise run against recordings). No need for stubbing as that is just a pain. We use node-tdd for thousands of tests and it is really well tested (and should be ok documented)

simlu:
updated

simlu:
Opened a ticked here regarding e2e for your consideration: https://github.com/public-transport/hafas-client/issues/148

derhuerst:
Merged! 🎉 Do you want to mentioned in the [`contributors` section](https://github.com/public-transport/hafas-client/blob/cf69ff4bc817bb45f721ed98663adefcebbab593/package.json#L17-L19)? If yes, how?

simlu:
@derhuerst All good. If I contribute more we can talk about it :)

PR #150, opened 2020-02-06
add VSN profile

n0emis:


p/vsn/index.js
```diff
@@ -0,0 +1,37 @@
...
+	transformReqBody,
+
+	products: products,
+
+	trip: true,
+	radar: true,
+	refreshJourney: true,
+	reachableFrom: true,
+
+	departuresGetPasslist: false,
```
derhuerst: Just asking to make sure: Did you copy-paste these flags, or did you check if the endpoint indeed doesn't support them?

derhuerst:
Thanks for your PR!

derhuerst:
If you have the resources, it would be great if you can add basic end-to-end tests covering the VSN endpoint.

Unfortunately, the E2E test code is not in a very good shape, and you need to copy-paste a lot of boilerplate. I suggest you to copy from the [`cfl` tests](https://github.com/public-transport/hafas-client/blob/master/test/e2e/hvv.js).

p/vsn/index.js
n0emis: ah, thanks for asking. Unfortunately I only partially tested the functionality; fixed with f6b694e

n0emis:
> Thanks for your PR! Do you want to be mentioned as a [collaborator in `package.json`](https://github.com/public-transport/hafas-client/blob/4837c2309ed0ac9f89dcf9724fcdc0234995eb85/package.json#L17-L19)? If yes, how?

yes, please: `n0emis <n0emis@noemis.me>`. Shall I add it myself?

derhuerst:
> yes, please: `n0emis <n0emis@noemis.me>`. Shall I add it myself?

Go ahead!

p/vsn/index.js
```diff
@@ -11,6 +11,21 @@ const transformReqBody = (ctx, body) => {
 	return body
 }
 
+const formatRefreshJourneyReq = (ctx, refreshToken) => {
+	const {profile, opt} = ctx
+
+	return {
+		meth: 'Reconstruction',
+		req: {
+			outReconL: [{ctx: refreshToken}],
```
derhuerst: It would be very interesting to know wether this is a change that all HAFAS endpoints with e.g. version `1.24`, or wether this is VSN-specific.

Do you know @marudor?

test/e2e/vsn.js
```diff
@@ -0,0 +1,242 @@
...
+const journeysFailsWithNoProduct = require('./lib/journeys-fails-with-no-product')
+const testDepartures = require('./lib/departures')
+const testArrivals = require('./lib/arrivals')
+
+const when = createWhen('Europe/Berlin', 'de-DE')
+
+const cfg = {
+	when,
+	products,
+	minLatitude: 47.24,
```
derhuerst: Adapt these to a reasonably large bounding box around the VSN service area. The intention here is to catch bogus values (e.g. `{latitude: 0, longitude: 0}` for VSN), so make it rather large.

test/e2e/vsn.js
```diff
@@ -0,0 +1,242 @@
...
+	maxLatitude: 52.9,
+	minLongitude: -0.63,
+	maxLongitude: 14.07
+}
+
+const _validateLine = createValidateLine(cfg)
+const validateLine = (validate, l, name) => {
+	if (!l.direction) l = Object.assign({}, l, {direction: 'foo'})
```
derhuerst: This code is in place because the CFL endpoint sometimes doesn't return a `direction`. Although I told you to copy the `cfl` tests, you can remove this code, unless you noticed that the `vsn` endpoint does the same.

Same with the `validateJourneyLeg` & `validateMovement` overrides.

test/e2e/vsn.js
```diff
@@ -0,0 +1,242 @@
...
+		validate,
+		fromId: kornmarkt,
+		toId: ewaldstrasse
+	})
+	t.end()
+})
+
+// todo: journeys, only one product
+
+test('journeys – fails with no product', (t) => {
```
derhuerst: Asserting this as part of the E2E tests doesn't make sense, it's a leftover of the crufty test code that I mentioned. You can remove it.

test/e2e/vsn.js
```diff
@@ -0,0 +1,191 @@
...
+const assert = require('assert')
+const tapePromise = require('tape-promise').default
+const tape = require('tape')
+const isRoughlyEqual = require('is-roughly-equal')
+
+const {createWhen} = require('./lib/util')
+const createClient = require('../..')
+const vsnProfile = require('../../p/vsn')
+const products = require('../../p/vsn/products')
+const {
```
derhuerst: Remove these, AFAICT they're not needed.

n0emis: hmm, without them, the code will not run...

derhuerst: But they're not used in the file.

derhuerst: I will remove this when merging.

derhuerst:
Merged and published as `hafas-client@5.1.0`. 🎉

PR #154, opened 2020-02-20
fix: performance improvements

simlu:


lib/scanner.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const assert = require('assert');
+const objectScan = require('object-scan');
+
+// Scan object and pre-compute results for provided needles (see object-scan for syntax)
+// Results are accessed per needle through callback function 'onResult'
+
+module.exports = (haystack, needles) => {
+	const precomputed = needles.reduce((p, c) => Object.assign(p, { [c]: [] }), {});
```
derhuerst: As this mutates `p` *anyways*, and as `precomputed` is local to the function, I prefer a for loop because of legibility.

Also, this uses `Object.create(null)` to allow for needles that clash with any field in the object prototype chain (e.g. `valueOf`).

```suggestion
	const precomputed = Object.create(null)
	for (const needle of needles) precomputed[needle] = []
```

simlu: I'm ok with doing a loop here. I love to chain functions, which is why I usually don't use them.

Regarding `Object.create(null)`: I'm really not sure how this could possible cause any issues. Everything seems perfectly safe when using `{}` instead of Object.create(null). Can you please explain?

simlu:
> > However currently the newly created branches are not scanned.
> 
> That's not necessary. The parsing logic doesn't add any _new_ references that need to be resolved.
> 
> I'm not really comfortable with the `Scanner(haystack, needles)(needle, onResult)` API though:
> 
> * Calling the inner function _looks_ like it would scan at that point, even though it has already scanned at the outer call.

I agree. I did this in anticipation that we might need to rescan the added branch. Since that is not necessary we can simplify.

> * Making the inner function throw when `needle` is not in `needles` is unintuitive.

It's an assert, not a throw. Ideally this could get removed in a compile step if so desired. 

`Assertions should only be used to verify conditions that should be logically impossible to be false (read: sanity checks).` [Reference](https://softwareengineering.stackexchange.com/questions/15515/when-to-use-assertions-and-when-to-use-exceptions)

I started really embracing asserts in the last few years since it makes debugging so much easier. However I agree that after refactoring this is no longer needed in this particular case.

> 
> Let's have a single function `scan` that returns an object `{[needle]: [{val, parent}]}`:
> 
> ```js
> const references = scan(res, {
> 	oprX: '**.oprX',
>  	icon: '**.icoX',
>  	prodX: '**.prodX',
>  	pRefL: '**.pRefL',
>  	locX: '**.locX',
> 	aniFLocX: '**.ani.fLocX',
> 	aniTLocX: '**.ani.tLocX',
> 	fLocX: '**.fLocX',
> 	tLocX: '**.tLocX',
> 	remX: '**.remX',
> 	himX: '**.himX',
> 	polyXL: '**.polyG.polyXL',
> })
> // …
> for (const {val: idx, parent}) {
> 	if ('number' === typeof idx) parent.operator = common.operators[idx]
> }
> // …
> ```

Agreed

simlu:
I've updated the PR, but did have a few comments. Please let me know what you think. Cheers!

lib/scanner.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const objectScan = require('object-scan');
+
+// Scan object and pre-compute results for provided needles (see object-scan for syntax)
+// Results are accessed per needle through callback function 'onResult'
```
simlu: Need to fix this comment

lib/scanner.js
simlu: If I understand correctly: If you forget to add a prototype needle you can still access this where the result is used without getting undefined. Is that what you meant?

lib/scanner.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const objectScan = require('object-scan');
+
+// Traverses object and returns object containing all matches of form:
+// { [needle]: [match, ...] } where match is [value, parents]
+// Refer to object-scan for needle syntax
```
simlu: Would you be ok with removing this comment entirely? I mean it's very obvious from how it's used and comment always get outdated (I comment very little these days and mostly on things that are not easily deduced from looking at the code).

simlu:
@derhuerst Ready for review again

lib/scanner.js
derhuerst: If you use a plain object with the default prototype chain, access using keys like `toString` & `valueOf` won't give `undefined`; `toString` is one of the more obvious cases where even authors of consuming code might be cautious, but there are more subtle cases.

To prevent this, one can either use a `Map`, or `Object.create(null)` to create an object without any prototype.

lib/scanner.js
derhuerst: Yes, it can be removed. The interesting part is possible `needle` syntaxes, and those shouldn't be documented here.

simlu: done

lib/scanner.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const objectScan = require('object-scan');
+
+module.exports = (haystack, needles) => {
```
derhuerst: Please give the function the name `findInTree`, for better stack trace readability.

simlu: done

PR #155, opened 2020-02-23
fix: improved benchmark test

simlu:


derhuerst:
👍 I merged a slightly modified version of your commit, which uses `lodash/cloneDeep` instead of `lodash.clonedeep`.

PR #157, opened 2020-02-27
add basic linter setup

juliuste:
I propose to add a very simple and unopinionated linter setup to the project to catch the most basic errors, like undefined- or unused variables.

Note that this setup excludes all tests and example files. IMHO this makes sense since things like unused variables are most likely intentional in tests while being accidental in the normal code base.

Also note that while I disabled the `no-irregular-whitespace` and `no-mixed-spaces-and-tabs` rules, it might actually make sense to enable them.

What's your opinion on this?

Also, to determine if this setup adds any value, here's the current linter error log for `master`:

> …/hafas-client/format/refresh-journey-req.js
>   4:9  error  'profile' is assigned a value but never used (no-unused-vars)
> 
> …/hafas-client/p/oebb/index.js
>   54:4  error  's' is constant (no-const-assign)
> 
> …/hafas-client/p/vsn/index.js
>   15:9  error  'profile' is assigned a value but never used (no-unused-vars)
> 
> …/hafas-client/parse/arrival-or-departure.js
>   3:7  error  'parseWhen' is assigned a value but never used     (no-unused-vars)
>   4:7  error  'parsePlatform' is assigned a value but never used (no-unused-vars)
> 
> …/hafas-client/parse/journey-leg.js
>   3:7  error  'parseWhen' is assigned a value but never used     (no-unused-vars)
>   4:7  error  'parsePlatform' is assigned a value but never used (no-unused-vars)
> 
> …/hafas-client/parse/location.js
>   30:20  error  'id' is not defined (no-undef)
>   30:67  error  'id' is not defined (no-undef)
> 
> …/hafas-client/parse/stopover.js
>   3:7  error  'parseWhen' is assigned a value but never used     (no-unused-vars)
>   4:7  error  'parsePlatform' is assigned a value but never used (no-unused-vars)
> 
> ✖ 11 problems (11 errors, 0 warnings)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ node_js:
   - 'lts/*'
   - '10'
 cache: npm
-script: npm run lint && npm test
+script:
+  - npm run lint
+  - npm test
 jobs:
   include:
     - name: e2e-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js:
   - 'lts/*'
   - '10'
 cache: npm
-script: npm test
+script: npm run lint && npm test
 jobs:
   include:
     - name: e2e-test
--- a/parse/location.js
+++ b/parse/location.js
@@ -27,7 +27,7 @@ const parseLocation = (ctx, l) => {
 		const stop = {
 			type: l.isMainMast ? 'station' : 'stop',
 			id: res.id,
-			name: l.name || id.O ? profile.parseStationName(ctx, l.name || id.O) : null,
+			name: l.name || lid.O ? profile.parseStationName(ctx, l.name || lid.O) : null,
 			location: 'number' === typeof res.latitude ? res : null // todo: remove `.id`
 		}
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,8 +14,6 @@
         "ecmaVersion": 2018
     },
     "rules": {
-        "no-irregular-whitespace": "off",
-        "no-mixed-spaces-and-tabs": "off",
         "no-unused-vars": [
             "error",
             {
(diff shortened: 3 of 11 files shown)


derhuerst:
> I propose to add a very simple and unopinionated linter setup to the project to catch the most basic errors, like undefined- or unused variables.
>
> Note that this setup excludes all tests and example files. [...]
>
> What's your opinion on this?

Sounds good!

I would like to keep unused variables in function signatures though; they help to understand the purpose of a function a lot, even if their body doesn't use them right now.

juliuste:
> I would like to keep unused variables in function signatures though; they help to understand the purpose of a function a lot, even if their body doesn't use them right now.

Should already be allowed, this is what the custom rule for `unused-vars` in the linter config does.

derhuerst:
Would you mind to fix the few linting problems? Then we can merge this. 🙌

parse/location.js
```diff
@@ -27,7 +27,7 @@ const parseLocation = (ctx, l) => {
 		const stop = {
 			type: l.isMainMast ? 'station' : 'stop',
 			id: res.id,
-			name: l.name || id.O ? profile.parseStationName(ctx, l.name || id.O) : null,
+			name: l.name || lid.O ? profile.parseStationName(ctx, l.name || lid.O) : null,
```
derhuerst: Nice! I had fixed that somewhere, but apparently not on `master`. 🙈

.travis.yml
```diff
@@ -5,7 +5,7 @@ node_js:
   - 'lts/*'
   - '10'
 cache: npm
-script: npm test
+script: npm run lint && npm test
```
derhuerst: This while give slightly cleaner CI output:

```suggestion
script:
  - npm run lint
  - npm test
```

PR #158, opened 2020-03-02
add VMT profile

derhuerst:


p/vmt/products.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
+
+module.exports = [ // todo: what is `512`?
+	{
+		id: 'long-distance-train',
+		mode: 'train',
+		bitmasks: [1, 2, 4],
+		name: 'long-distance train',
```
juliuste: Stupid little question: Do we capitalize the names here (as in INVG) or not? We should be consistent.

derhuerst:
> What about e2e-tests?

I will add some basic ones. Given that we have some "unit tests" now, and that we know the differences between endpoints better by now, IMO we don't need that many E2E tests anymore.

p/vmt/products.js
derhuerst: I don't know. These descriptions are not internationalised anyways, so I rather see them as useful for prototype-level apps and as a fallback.

PR #159, opened 2020-03-02
add RSAG profile

derhuerst:


p/rsag/example.js
```diff
@@ -0,0 +1,43 @@
+'use strict'
+
+const createClient = require('../..')
+const hvvProfile = require('.')
+
+const client = createClient(hvvProfile, 'hafas-client-example')
```
juliuste:
```suggestion
const rsagProfile = require('.')

const client = createClient(rsagProfile, 'hafas-client-example')
```

p/rsag/products.js
```diff
@@ -0,0 +1,84 @@
...
+		bitmasks: [4],
+		name: 'InterRegio/high-speed train',
+		short: 'IR/other',
+		default: true
+	},
+	{
+		id: 'regional-train', // todo: rename
+		mode: 'train',
+		bitmasks: [8],
+		name: 'regional train',
```
juliuste: Do we capitalize these names?

PR #160, opened 2020-03-03
add RMV profile

derhuerst:


p/rmv/example.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+
+const createClient = require('../..')
+const hvvProfile = require('.')
+
+const client = createClient(hvvProfile, 'hafas-client-example')
```
juliuste:
```suggestion
const rmvProfile = require('.')

const client = createClient(rmvProfile, 'hafas-client-example')
```

test/e2e/rmv.js
```diff
@@ -0,0 +1,114 @@
...
+	validate(t, trip, 'trip', 'trip')
+	t.end()
+})
+
+test('arrivals at Wiesbaden Hbf', async (t) => {
+	const arrivals = await client.arrivals(wiesbadenHbf, {
+		duration: 10, when
+	})
+
+	validate(t, arrivals, 'arrivals', 'arrivals')
```
juliuste: Is there any reason you don't use the `testArrivals` method that's imported above here?

derhuerst: `testArrivals` asserts that the resulting arrivals have the `stop.id`/`stop.station.id` that we passed in as `id`. Some endpoints, such as the RMV one, use varying stop IDs though.

derhuerst: The test tooling is bad right now, will eventually™ improve it.

PR #161, opened 2020-03-03
add VBN profile

derhuerst:


p/vbn/products.js
```diff
@@ -0,0 +1,84 @@
...
+	{
+		id: 'national-train',
+		mode: 'train',
+		bitmasks: [2],
+		name: 'InterCity, EuroCity, CityNightLine',
+		short: 'IC/EC/CNL',
+		default: true
+	},
+	{
+		id: 'national-train-2',
```
juliuste: I think this id is sub-optimal. We should either group national trains or give this one a different id.

PR #166, opened 2020-03-11
add PKP profile

juliuste:
still missing: e2e tests


p/pkp/index.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+const products = require('./products')
+
+const transformReqBody = (ctx, body) => {
+	body.client = {type: 'AND', id: 'HAFAS', name: 'Rozklad-PKP', v: '1000110'}
+	body.ver = '1.18' // 1.24 is used by the app, but that version doesn't support getPassList for departures/arrivals
```
derhuerst: I *think* you can use `1.20`. Not sure if there are differences, but I'd err on the side of more features.

juliuste: No, same issue with `1.20`:

```
{ [Error: Parser error: root.svcReqL.svcReqL.req(getPasslist)] isHafasError: true, request: '{"lang":"en","svcReqL":[{"meth":"StationBoard","req":{"type":"ARR","date":"20200316","time":"100000","stbLoc":{"type":"S","lid":"A=1@L=5100028@"},"dirLoc":null,"jnyFltrL":[{"type":"PROD","mode":"INC","value":"47"}],"dur":10,"getPasslist":false,"stbFltrEquiv":false}}],"client":{"type":"AND","id":"HAFAS","name":"Rozklad-PKP","v":"1000110"},"ver":"1.20","auth":{"type":"AID","aid":"DrxJYtYZQpEBCtcb"}}', url: 'https://mobil.rozklad-pkp.pl:8019/bin/mgate.exe?', statusCode: 200, code: 'PARSE', message: 'Parser error: root.svcReqL.svcReqL.req(getPasslist)', responseId: null }
```

PR #176, opened 2020-04-23
add a TypeScript Declaration File

bergmannjg:
This PR adds a TypeScript Declaration File for the hafas-client api.

You can use it in TypeScript  and other languages.

This [F#/Fable App](https://github.com/bergmannjg/hafas-client-fable) is an example use case. There is a script to test the coverage of the declaration file with the hafas-client interface.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,314 @@
+// typescript declaration file for hafas-client, https://www.npmjs.com/package/hafas-client
+
+// info: Leg, Trip and StopOver have no type members
+
+export interface Location {
+    type: 'location',
+    id?: string,
+    name?: string,
+    poi?: boolean,
+    address?: string,
+    longitude?: number,
+    latitude?: number,
+    altitude?: number
+}
+
+export interface Products {
+    nationalExpress?: boolean,
+    national?: boolean,
+    regionalExp?: boolean,
+    regional?: boolean,
+    suburban?: boolean,
+    bus?: boolean,
+    express?: boolean,
+    ferry?: boolean,
+    subway?: boolean,
+    tram?: boolean,
+    taxi?: boolean,
+    "high-speed-train"?: boolean,
+    "intercity-p"?: boolean,
+    "local-train"?: boolean,
+    metro?: boolean,
+    "s-train"?: boolean,
+}
+
+export interface Station {
+    type: 'station',
+    id: string,
+    name: string,
+    station?: Station,
+    location?: Location,
+    products?: Products,
+    regions?: string[] // region ids
+}
+
+export interface Stop {
+    type: 'stop',
+    id: string,
+    station?: Station,
+    name: string,
+    location?: Location,
+    products: Products
+}
+
+export interface Region {
+    type: 'region',
+    id: string,
+    name: string,
+    stations: string[] // station ids
+}
+
+export interface Line {
+    type: 'line',
+    id: string,
+    name: string,
+    adminCode?: string,
+    fahrtNr?: string,
+    additionalName?: string,
+    product?: string,
+    public?: boolean,
+    mode: 'train' | 'bus' | 'watercraft' | 'taxi' | 'gondola' | 'aircraft' | 'car' | 'bicycle' | 'walking',
+    routes?: string[] // routes ids
+    operator?: Operator
+}
+
+export interface Route {
+    type: 'route',
+    id: string,
+    line: string,
+    mode: 'train' | 'bus' | 'watercraft' | 'taxi' | 'gondola' | 'aircraft' | 'car' | 'bicycle' | 'walking',
+    stops: string[] // stop ids
+}
+
+export interface ArrivalDeparture {
+    arrival?: number,
+    departure?: number
+}
+
+export interface Schedule {
+    type: 'schedule',
+    id: string,
+    route: string
+    mode: 'train' | 'bus' | 'watercraft' | 'taxi' | 'gondola' | 'aircraft' | 'car' | 'bicycle' | 'walking',
+    sequence: ArrivalDeparture[],
+    starts: string[] // array of Unix timestamps
+}
+
+export interface Operator {
+    type: 'operator',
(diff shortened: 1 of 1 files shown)


derhuerst:
Hey, thanks for contributing!

> This F#/Fable App is an example use case.

It's exciting to see people build things using `hafas-client` with *other* languages than JavaScript; Hopefully we can overcome the usual level of [*not invented here*](https://en.wikipedia.org/wiki/Not_invented_here) in this specific niche.

I'm not that comfortable with maintaining Typescript typings though.

I see the following pros:

- improved DX when working with TypeScript and TypeScript-compatible tools
- a stronger statement that `hafas-client` intends not to break its API, except with major releases

I see the following cons:

- I don't use TypeScript, so maintaining the typings is additional overhead for me. I even think it's likely that the actual JS code and the typings will diverge because I forget about them; This might be fine if we keep them very shallow.
- For some people, the typings may induce a false sense of type safety; Underneath, `hafas-client` is still loosely typed.
- This is a minor point, but I dislike the idea of maintaining code that could be generated using static analysis to some extent.

In addition, the typings in this PR are simpler than they would actually have to be:

- [`hafas-client` has one *profile* for each HAFAS endpoint it supports](https://github.com/public-transport/hafas-client/blob/e02a20b1de59bda3cd380445b6105e4c46036636/p/readme.md). Each profile defines its own list of products, as they reflect the data returned by the corresponding endpoint. `Procucts` therefore depends on the profile being used.
- A profile might support more options than specific in e.g. `JourneysOptions`, but a `*Options` type sounds like it's exhaustive. For example, the [`db` profile](https://github.com/public-transport/hafas-client/tree/e02a20b1de59bda3cd380445b6105e4c46036636/p/db) additionally supports `firstBlass` & `loyaltyCard`.

bergmannjg:
Thank you for the quick answer,

i think, i underestimated your additional overhead to keep the types in sync with the api.

In addition to the F# App i have a TypeScript App using hafas-client, and the ts declaration file
helped me alot to find bugs.

I want to share my findings, maybe its a good alternative that i make a new repo for the ts declaration file and clarfify the restrictions. I can record the coverage of the api with the F# App.

So everybody can use it and i can integrate it in my build pipelines.

derhuerst:
If you'll be managing TypeScript typings *anyways*, please do so in the [DefinitelyTyped project](https://github.com/DefinitelyTyped/DefinitelyTyped). AFAIK a lot of TS-related tooling picks up definitions from there automatically.

bergmannjg:
thank you for the tip, i will publish the  declaration file in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

derhuerst:
FYI, in case you want to keep maintaining them: You can watch only the releases in this repo.

bergmannjg:
The declaration file is merged in DefinitelyTyped and [npm](https://www.npmjs.com/package/@types/hafas-client).

bergmannjg:
I've made a [hafas-rest-api](https://github.com/bergmannjg/hafas-api-tsoa) in TypeScript using the [Tsoa](https://github.com/lukeautry/tsoa) Framework. It's like a proof of concept.

It generates an OpenAPI spec of the Api, there is a Web UI for documentation and api execution and to generate clients in many programming languages.

The connection to hafas-client is in this [controller](https://github.com/bergmannjg/hafas-api-tsoa/blob/master/src/hafas/hafasController.ts) class. The api call and the parameters with defaults and comments are in the same function.  

As a step further it may be possible to generate the whole function from the interfaces in 
the ts declararation file.

@derhuerst what do you think about this.

derhuerst:
Thanks again for your efforts!

Recently, I [started letting `hafas-rest-api` generate API docs from a custom API spec](https://github.com/public-transport/hafas-rest-api/blob/1c4dcc4c6b8b89a087440851088fbc2eb0b5b363/tools/generate-docs.js). This custom spec is obviously not a proper solution, just a convenient "local optimum" given the current `hafas-client` & `hafas-rest-api` implementations!

Implementing them, I had similar wishes like you: **A machine-readable spec of the `hafas-client` API, as reusable as possible, to generate downstream code and/or tools from**. @juliuste also brought this up today, mentioning that he would like to improve the interoperability of `hafas-client` to other languages (e.g. Python, which is currently only "dealt with" by a [JSON-RPC bridge](https://www.npmjs.com/package/hafas-client-rpc)). So we're three people with significant demand for this already!

---

Given that TypeScript (TS) is a natural choice for JS projects, I'll sum up possible changes:

- Re-implementing `hafas-client` in TS is most likely not going to be worth it:
	- We'd have typings *anyways*, which we could reuse for the use cases above.
	- My TypeScript knowledge & experience is limited, but due to [the highly dynamic way `hafas-client` profiles work](https://github.com/public-transport/hafas-client/blob/156de8e1c529c8eb11a170c2ed6772d00cef487f/docs/writing-a-profile.md) ([example](https://github.com/public-transport/hafas-client/blob/156de8e1c529c8eb11a170c2ed6772d00cef487f/docs/migrating-to-4.md#if-you-use-lineclass-or-lineproductcode)), typing would likely be very complex (and therefore make development slower), and hard to actually make use of in downstream tools.
	- Using TypeScript would give use *only some* of the safety that it is usually praised for: It may bugs within `hafas-client` (e.g. code expecting a property not to be `null`), but HAFAS can still leave out pieces of the response "tree" (e.g. a station ID), so almost all fields would have to be "optionals".
- Adding TS typings might be worth it:
	- They would enable the use cases above.
	- They would probably diverge from the actual JS implementation, but that would IMHO be negligible, given that HAFAS itself is highly unreliable *anyways*.
	- They would not slow down the development & maintenance of `hafas-client`.
- Adding a different typing system, e.g. [JSON schema](https://json-schema.org).
	- *benefits & downsides just like with TS typings*

JSON schema specs probably have a wider reach beyond the JS/TS world, but given that TS has become very popular recently, both should probably be fine.

PR #177, opened 2020-05-21
integration tests using recorded HTTP requests

### LINKED ISSUE #148: Consider using nock for e2e tests
Consider using [nockBack](https://github.com/nock/nock#readme) for e2e test. We a very convenient wrapper for mocha called [node-tdd](https://github.com/blackflux/node-tdd#usenock).

The way this would work:
1. Run tests against external apis as necessary, nockBack / node-tdd automatically captures requests and stores them in cassettes
2. Re-running the tests will not make external requests. Instead the request is intercepted and the recording is returned. So if the external service is temporarily unavailable or slow, this wont impact test success. No external requests should ever be made when test are run as part of a suite
3. When a service changes their API, delete the recordings for the relevant tests and re-run the tests (fix code as necessary). Then commit the new cassettes

This is a solid approach when dealing with external APIs and considered best practice.

derhuerst:
closes #148

to do:

1. [x] adapt E2E tests not to fail
2. [x] check in final `test/e2e/fixtures`


test/e2e/lib/util.js
```diff
@@ -20,9 +23,32 @@ const assertValidWhen = (actual, expected, name) => {
...
+	replayer.configure({
+		headerWhitelist: [
+			'Content-Type', 'Accept-Encoding', 'Accept',
+		],
+		includeHeaderValues: true,
+		touchHits: false,
+	})
+	replayer.fixtureDir(join(__dirname, '..', 'fixtures'))
+}
+const test = tape
```
juliuste, line 49: Shouldn't we still use `tape-promise` here, since requests are technically still async and one could disable `process.env.VCR_MODE`?

derhuerst: [`tape@5` supports async test functions.](https://github.com/substack/tape/blob/3554d4b049f34065c6a643c5cda3d88c70a4908e/lib/test.js).

juliuste: ah, nevermind then.

Issue #179, opened 2020-06-09
Direction field for arrivals is always null


marzieh-kazemi:
The `direction` field for arrivals return always null for any trip. @derhuerst @juliuste :)

The version we use => ^5.0.2

<img width="635" alt="Screenshot 2020-06-08 at 12 44 34 PM" src="https://user-images.githubusercontent.com/41323375/84127864-0f53cb00-aa40-11ea-811e-b0eabee73679.png">

derhuerst:
This is due to 044a5ee816d1ba97beef0687bc38b1e5e5c53b94. HAFAS does not seem to return the direction (as in where the trip will end) for arrivals, but instread just where the trip started.

We should expose that information though. PR welcome!

marzieh-kazemi:
But we don't get where the trip started, I thought it will be covered by direction like departure? What is the field for it then?

juliuste:
I'd suggest adding a field `provenance` for arrivals then, which contains this value. This is very useful to have e.g. for an departures/arrivals board.

PR #180, opened 2020-06-09
add provenance field on arrivals

juliuste:
As discussed in #179, the `direction` field on `arrival`s doesn't contain the place the vehicle is going *towards*, but rather where it came from. We therefore currently discard this information on arrivals, but IMHO it would still make sense to expose it, just under a different name. 

I suggest `provenance` for that, but feel free to make other proposals.


derhuerst:
It would be great to have a test similar to `test/vbb-departures.js` for this, e.g. `test/db-arrivals.js`.

juliuste:
done, please review again

PR #187, opened 2020-08-01
Fix platform parsing with some profiles

em0lar:
With some profiles hafas-client is unable to parse platforms. This happens with departures/arrival requests, as well as with everywhere where journey-leg or stopover parsing is used. The profiles where I saw this behavior are VSN and ZVV, but I only tried BVG and DB, so there can be more profiles with this issue.

As I'm developer of pyHaFAS I saw that there a multiple ways HaFAS can return the platform. Here is an example for data that hafas-client can accept.
```json
{
  "dPlatfS": "10 A-B",
  "dPlatfR": "9",
}
```

Here are two examples for data returned from the VSN-HaFAS:
```json
{
  "dPltfS": { "type": "PL", "txt": "E" },
}

{
  "dPltfS": { "type": "PL", "txt": "10" },
  "dPltfR": { "type": "U", "txt": "10" }
}
```

I'm not sure what the `type` means, so this PR only covers the `txt` attribute. This attribute is now used as alternative when `XPlatfY` is not available.

I'm open to any recommends on code style (or anything else), since I'm not that experienced with JS.

### DIFF
--- a/parse/arrival-or-departure.js
+++ b/parse/arrival-or-departure.js
@@ -19,8 +19,8 @@ const createParseArrOrDep = (prefix) => {
 		const tPrognosed = d.stbStop[prefix + 'TimeR']
 		const tzOffset = d.stbStop[prefix + 'TZOffset'] || null
 		const cancelled = !!d.stbStop[prefix + 'Cncl']
-		const plPlanned = d.stbStop[prefix + 'PlatfS']
-		const plPrognosed = d.stbStop[prefix + 'PlatfR']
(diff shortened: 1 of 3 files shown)


derhuerst:
Thanks for this PR!

Do you have a specific request where this happens? I'd like to add the (full) response as a test case.

derhuerst:
In 2612494, I have pushed your commit, with small style modifications. In da01544, I have added a test case.

Do you want to be [mentioned](https://github.com/public-transport/hafas-client/blob/da01544b42aa02bef57906ad5ef4f3e1b456732e/package.json#L17-L20) as a contributor? If yes, with which name?

em0lar:
> Do you want to be mentioned as a contributor? If yes, with which name?

Yes, please: `em0lar <hello@em0lar.de>`

derhuerst:
Published as `hafas-client@5.7.1`. 🎉

PR #188, opened 2020-08-10
add Irish Rail profile

derhuerst:


p/irish-rail/index.js
```diff
@@ -0,0 +1,36 @@
...
+		type: 'IPA',
+		id: 'IRISHRAIL',
+		v: '4000100',
+		name: 'IrishRailPROD-APPSTORE',
+		os: 'iOS 12.4.8',
+	}
+	body.ver = '1.18'
+
+	body.auth = {type: 'AID', aid: 'P9bplgVCGnozdgQE'}
+	body.lang = 'de'
```
juliuste, line 16: Is this intentional?

derhuerst: Nope. This should probably be `en`, right?

juliuste: Yes, or `ga`, not sure.

PR #198, opened 2020-11-18
WIP: SBB profile

### LINKED ISSUE #197: add SBB profile (swiss rail)
SBB operates a HAFAS endpoint: http://fahrplan.sbb.ch/bin/mgate.exe

Sidenote: SBB leads by example and provides [open transit data](https://opentransportdata.swiss/de/dataset). Still it may be useful to query their HAFAS

noerw:
closes #197 

still WIP, todo:

- [x] think about locale: switzerland is multilingual, should we hardcode `de_CH`?
- [x] extract products with mitmproxy & SBB Mobile
- [ ] fix issues with `journeys()`:
  - sometimes it works
  - sometimes throws
    ```js
    {
      isHafasError: true,
      request: '{"lang":"deu","svcReqL":[{"cfg":{"polyEnc":"GPA"},"meth":"TripSearch","req":{"getPasslist":false,"maxChg":-1,"minChgTime":0,"depLocL":[{"type":"S","lid":"A=1@L=8590197@"}],"viaLocL":null,"arrLocL":[{"type":"S","lid":"A=1@L=8592321@"}],"jnyFltrL":[{"type":"PROD","mode":"INC","value":"7167"},{"type":"META","mode":"INC","meta":"notBarrierfree"}],"gisFltrL":[],"getTariff":false,"ushrp":true,"getPT":true,"getIV":false,"getPolyline":false,"outDate":"20201118","outTime":"134835","outFrwd":true}}],"client":{"type":"AND","id":"DBZUGRADARNETZ","v":""},"ver":"1.18","auth":{"aid":"hf7mcf9bv3nv8g5f","type":"AID"}}',
      url: 'http://fahrplan.sbb.ch/bin/mgate.exe?',
      statusCode: 200,
      code: 'CGI_READ_FAILED',
      message: 'CGI_READ_FAILED',
      responseId: null
    }
    ```
  - sometimes throws
    ```
    TypeError: Cannot read property '0' of undefined
      at parseAlternative (/home/norwin/src/hafas-client/parse/journey-leg.js:143:24)
      at Array.map (<anonymous>)
      at Object.parseJourneyLeg (/home/norwin/src/hafas-client/parse/journey-leg.js:151:33)
    ```

### DIFF
--- a/p/sbb/index.js
+++ b/p/sbb/index.js
@@ -1,0 +1,32 @@
+'use strict'
+
+const products = require('./products')
+
+const transformReqBody = (ctx, body) => {
+	// sourced from https://github.com/marudor/BahnhofsAbfahrten/blob/1c99d8b/packages/server/HAFAS/profiles.ts#L344-L361
+	body.client = {
+		type: 'AND',
+		id: 'DBZUGRADARNETZ',
+		v: '',
+	}
+	body.ver = '1.18'
+	body.auth = { aid: 'hf7mcf9bv3nv8g5f', type:'AID' }
+	body.lang = 'deu'
+
+	return body
+}
+
+const sbbProfile = {
+	locale: 'de-CH',
+	timezone: 'Europe/Zurich',
+	endpoint: 'http://fahrplan.sbb.ch/bin/mgate.exe',
+	transformReqBody,
+
+	products: products,
+
+	trip: true,
+	radar: true,
+	reachableFrom: true,
+}
+
+module.exports = sbbProfile
--- a/p/sbb/readme.md
+++ b/p/sbb/readme.md
@@ -1,0 +1,16 @@
+# SBB profile for `hafas-client`
+
+[*Schweizerische Bundesbahnen (SBB)*](https://en.wikipedia.org/wiki/Swiss_Federal_Railways) is the largest Swiss long-distance public transport company. This profile adds *SBB*-specific customizations to `hafas-client`.
+
+## Usage
+
+```js
+const createClient = require('hafas-client')
+const sbbProfile = require('hafas-client/p/sbb')
+
+// create a client with ÖBB profile
+const client = createClient(sbbProfile, 'my-awesome-program')
+```
+
+
+## Customisations
--- a/p/sbb/example.js
+++ b/p/sbb/example.js
@@ -1,0 +1,50 @@
+'use strict'
+
+const createClient = require('../../')
+const sbbProfile = require('.')
+
+const client = createClient(sbbProfile, 'hafas-client-example')
+
+const zurich1 = '8530813' // Zürich Kreuzplatz
+const zurich2 = '8503000' // Zürich HB
+const basel1 = '8500073'  // Basel Aeschenplatz
+const basel2 = '8500010'  // Basel SBB
+const baden1 = '8590197'
+const baden2 = '8590173'
+const locBaden = { type: 'location', latitude: 47.476, longitude: 8.30613 }
+
+// FIXME: journeys() sometimes fails with CGI_READ_FAILED.
+// FIXME: journeys() sometimes fails with    TypeError: Cannot read property '0' of undefined
+//     at parseAlternative (/home/norwin/src/hafas-client/parse/journey-leg.js:143:24)
+
+// client.journeys(zurich1, basel1, { results: 1 })
+// client.journeys(zurich2, basel2, { results: 1 })
+client.journeys(baden1, baden2, { results: 1 })
+// client.departures(zurich2, {duration: 1})
+// client.arrivals(zurich2, {duration: 10, linesOfStops: true})
+// client.locations('Basel', { results: 2 })
+// client.locations('Luzern', { poi: true, stops: false })
+// client.stop(basel2)
+// client.nearby(locBaden, { distance: 100 })
+// client.radar({
+// 	north: 47.5,
+// 	west: 8.2,
+// 	south: 47.3,
+// 	east: 8.5,
+// }, { results: 10 })
+// client.reachableFrom({
+// 	type: 'location',
+// 	id: '980023031',
+// 	latitude: 47.052417,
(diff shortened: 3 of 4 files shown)


noerw:
Also: I saw that other profiles implement various workarounds. Is there some testsuite to verify the new profile works as expected or needs some workarounds?

derhuerst:
Thanks for the PR! 👏

> think about locale: switzerland is multilingual, should we hardcode `de_CH`?

Not sure, I had the same thought with the SNCB/NMBS profile.

Allowing people to pass in their preferred locale via an additional parameter would introduce an exception just for 2 profiles right now, so dislike it a lot. What about documenting in `p/sbb/readme.md` how to adapt the locale?

```js
const createClient = require('hafas-client')
const sbbProfile = require('hafas-client/p/sbb')

const client = createClient({
    ...sbbProfile, locale: 'fr_CH',
}, 'some SBB client')
```

> extract products with mitmproxy & SBB Mobile

The [*writing a profile* guide](https://github.com/public-transport/hafas-client/blob/5/docs/writing-a-profile.md) may help you with that.

> fix issues with `journeys()`: sometimes it works, sometimes throws `CGI_READ_FAILED`

You're very welcome to investigate! With server-side errors such as this one, I couldn't always find a cause in the past, so don't waste too much time.

> sometimes throws: `Cannot read property '0' of undefined`

With a request that provokes this error, can you run with the `DEBUG=hafas-client` environment variable and paste the response body here? That would help me fix it.

p/sbb/index.js
```diff
@@ -0,0 +1,32 @@
...
+	body.auth = { aid: 'hf7mcf9bv3nv8g5f', type:'AID' }
+	body.lang = 'deu'
+
+	return body
+}
+
+const sbbProfile = {
+	locale: 'de-CH',
+	timezone: 'Europe/Zurich',
+	endpoint: 'http://fahrplan.sbb.ch/bin/mgate.exe',
```
derhuerst, line 22: The `https:` endpoint actually redirects all requests to `http:`. 😢

p/sbb/readme.md
```diff
@@ -0,0 +1,16 @@
...
+## Usage
+
+```js
+const createClient = require('hafas-client')
+const sbbProfile = require('hafas-client/p/sbb')
+
+// create a client with ÖBB profile
+const client = createClient(sbbProfile, 'my-awesome-program')
+```
+
```
derhuerst, line 14: Let's add a note about how to customize `profile.locale` & `lang` here.

derhuerst:
> Also: I saw that other profiles implement various workarounds. Is there some testsuite to verify the new profile works as expected or needs some workarounds?

The `hafas-client` tests intend to answer *two* questions:

- **Does it work as intended with a (live) endpoint?** For this, there should be "end-to-end" tests querying the SBB endpoint for real connections/departures/etc and checking that they make sense. The [`hvv` E2E tests implement this in a minimal but sufficient way](https://github.com/public-transport/hafas-client/blob/0e46be72372ead18bc6758cb54223998eb86e031/test/e2e/hvv.js) (sorry, the test setup is very manual and verbose), other profiles have additional tests for special features & behaviours, [e.g. `bvg` with *BerlKönig*](https://github.com/public-transport/hafas-client/blob/0e46be72372ead18bc6758cb54223998eb86e031/test/e2e/bvg.js#L134-L167). I'll try to update these tests to the changing operating situation from time to time, e.g. discontinued trains.
- Working on `hafas-client`, **did I introduce a regression?** For this, HAFAS responses to the E2E tests are checked into Git, and then run in an entirely reproducible way. This way, we only test if `hafas-client` still a) produces the exact same requests and b) parses them properly. Please follow the [*adding integration/end-to-end tests* guide](https://github.com/public-transport/hafas-client/blob/0e46be72372ead18bc6758cb54223998eb86e031/contributing.md#adding-integrationend-to-end-tests).

p/sbb/index.js
noerw: :( I guess we should hardcode `https://` then, in the hopes that this is changed at some point

p/sbb/readme.md
noerw: what do you think of implementing something like
```
const profile = require('hafas-client/p/sbb').forLocale(locale: 'fr_CH' | 'it_CH' | 'de_CH')
```

noerw:
> The hafas-client tests intend to answer two questions:

iow, systematic tests for semantically meaningful responses do not exist? Then I'll have to do some manual sanity checks by "random" sampling. Though I have no local knowledge 🙃

I'll implement those sampels as E2E tests then

p/sbb/index.js
derhuerst: Please try if the `fetch` API then re-sends the body for the 2nd request; If it doesn't, there's probably a good reason not to send the body after a redirect.

p/sbb/readme.md
derhuerst: I like that, from a user perspective, it's a reasonably ergonomic API, and it doesn't interfere with the `createClient` logic in `hafas-client/index.js`. I don't like though that `forLocale` effectively becomes part of the profile passed into `createClient`.

What about having two (three if you include Italian) entrypoints `hafas-client/p/sbb/de-ch` & `hafas-client/p/sbb/fr-ch`? There would be a additional file, but it would just extend `require('./de-ch')`.

noerw: nice and simple, i like it

derhuerst:
> iow, systematic tests for semantically meaningful responses do not exist?

I don't think so.

> Then I'll have to do some manual sanity checks by "random" sampling. Though I have no local knowledge 🙃

Yeah, I did they same: picked random places in different cities.

derhuerst:
Are you still willing to work on this? Otherwise, I'd continue.

noerw:
@derhuerst Sorry, I forgot about this & currently don't have the time to continue, so go ahead if you want :)

PR #210, opened 2021-01-19
update profiles

derhuerst:
still to do:

- [x] add `departuresGetPasslist: false` & `departuresStbFltrEquiv: false` for all `ver >= 1.20` profiles
- [x] decide if this ^ is a breaking change – I decided that it's not. Old versions that still support these ^ will be shut off eventually.
- [x] update `p/*/example.js`


derhuerst:
@juliuste Please take a quick look at the changed code. You can ignore the CI status, 5 E2E/integration tests are still failing, we can't fix them easily, but they're not significant.

PR #223, opened 2021-03-25
Enable usage of 1st class BahnCards and fix typo

### LINKED ISSUE #221: BahnCard 1st class not selectable
Hey there,

In https://github.com/public-transport/hafas-client/blob/68d8bf9fca31de0608b5ba7b7dbdb57a96a33a59/p/db/loyalty-cards.js#L17-L18 `c.class` always seems to be `undefined`, which means that the BahnCard 2nd class is always used. This is problematic because it means there is no BahnCard discount when booking 1st class trips.

I would suggest to either replace `c.class` with `data.class` which allows the user to choose the class of the BahnCard or use `firstClass` from `opt` to provide information about class.

Also https://github.com/public-transport/hafas-client/blob/68d8bf9fca31de0608b5ba7b7dbdb57a96a33a59/p/db/loyalty-cards.js#L5 contains a typo (`loyaly`).

Thank you in advance!

roehrt:
Resolves #221

### DIFF
--- a/p/db/loyalty-cards.js
+++ b/p/db/loyalty-cards.js
@@ -2,7 +2,7 @@
 
 // todo: generate from https://reiseauskunft.bahn.de/addons/fachkonfig-utf8.cfg ?
 const c = {
-	NONE: Symbol('no loyaly card'),
+	NONE: Symbol('no loyalty card'),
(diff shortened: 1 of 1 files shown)


derhuerst:
Published dbac565 as `hafas-client@5.15.2`. 🎉

PR #224, opened 2021-03-25
Replace the 'source code' column with a 'profile name' column

KristjanESPERANTO:
This is helpful to choose the right profile. The folder name corresponds to the 'profile id' that I have to specify when I create a client. I haven't seen such an overview yet.

I'm not sure if 'profile id' is the appropriate term.


derhuerst:
I'd prefer to call it "profile name", because sounds more technical than it actually is IMHO.

Also, what do you think about just replacing the "source code" column with the profile name/id column? Most people won't be interested in the source code of the profile, and if they are, they can still open `index.js`.

KristjanESPERANTO:
Yes good idea. I have now adjusted it as I understand it. Is this okay?

derhuerst:
Yes, looks good!

Optional cherry on top: Mark the profile name/ID as code, because it's casing is relevant, as it used to `require()` the profile.

KristjanESPERANTO:
I changed it :slightly_smiling_face:

derhuerst:
Do you want to be mentioned as a contributor? If yes, how? Just as `Kristian`, or with an email address or a URL?

KristjanESPERANTO:
I actually feel that my contribution is too small to be mentioned. But if you'd still like to mention me, I'd like that. Then with `Kristjan Esperanto` and a link to my GitHub profile :slightly_smiling_face:

derhuerst:
Published dbac565 as `hafas-client@5.15.2`. 🎉

PR #240, opened 2021-08-17
Expose `realtimeDataFrom` for refreshJourney

### LINKED ISSUE #239: Expose `realtimeDataFrom` for refreshJourney
`journeys()` returns a `realtimeDataFrom`, but `refreshJourney()` does not, even though the data is in the raw response

yu-re-ka:
Fixes #239


derhuerst:

PR #241, opened 2021-08-18
fix db price parsing

yu-re-ka:
Unfortunately DB does not return price information for refreshJourney, even when `tickets: true`.

### DIFF
--- a/p/db/index.js
+++ b/p/db/index.js
@@ -186,14 +186,26 @@ const parseJourneyWithPrice = ({parsed}, raw) => {
 	parsed.price = null
 	// todo: find cheapest, find discounts
 	// todo: write a parser like vbb-parse-ticket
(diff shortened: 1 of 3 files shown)


derhuerst:
I can confirm that the DB endpoint returns the new price information.

Can you adapt the unit test fixtures? You can check if the unit tests are running with `npm run test-unit`.

yu-re-ka:
Should I just change the existing fixtures to match the new price formatting or create new ones? If I should create new ones, how do I find a journey with a DEVI leg?

derhuerst:
Just adapt the fixture.

yu-re-ka:
done

derhuerst:
Thanks!

derhuerst:

PR #244, opened 2021-08-20
expose prodCtx.catOut as line.productName

### LINKED ISSUE #242: Provide prodCtx.catOut field
I would call it `line.productName` in the output. It contains something like `"ICE", "IC", "S", "U", "Tram", "Bus"`, but also sometimes things that are not part of the profile product map like `"ECE", "EST", "EN", "NJ", "HLB", "BRB"`. This is useful for displaying which products are used in a connection. It also seems to be consistent across different profiles.

screenshot1

On the DB profile I can split off (with space as delimiter) the first part of line.name and it is the same as `catOut`, but that seems like a hack and it doesn't work on other profiles like VBB and BVG.

yu-re-ka:
Closes #242

### DIFF
--- a/parse/line.js
+++ b/parse/line.js
@@ -25,6 +25,13 @@ const parseLine = ({profile}, p) => {
 		res.adminCode = p.prodCtx.admin.replace(/-+$/, '')
 	}
 
+	if (p.prodCtx && 'string' === typeof p.prodCtx.catOut) {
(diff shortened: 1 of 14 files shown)


derhuerst (approved):

derhuerst:
Thanks!

derhuerst:

PR #249, opened 2021-11-01
request.js: Use correct HttpsAgent option for localAddress

MTRNord:
This most likely has been a typo when it was added but according to the [node docs](https://nodejs.org/api/net.html#socketconnectoptions-connectlistener) this field is actually called `localAddress`.

This change also made this locally work again after one of my IPs got blocked by a hafas endpoint.


derhuerst:
Thanks for pointing this out! Huh, I'm pretty sure that I validated this code against a server while ago. 🤔

derhuerst:
Published as `hafas-client@5.21.0`. 🎉

PR #251, opened 2021-11-27
add KVB profile

derhuerst:


derhuerst:
Published as `hafas-client@5.22.0`. 🎉

PR #278, opened 2022-11-20
dbProfile -> profile

KristjanESPERANTO:
The examples in the docs don't work because `profile` is returned instead of `dbProfile`.

### DIFF
--- a/docs/departures.md
+++ b/docs/departures.md
@@ -44,9 +44,9 @@ If you pass an object `opt.products`, its fields will partially override the def
 
 ```js
 import {createClient} from 'hafas-client'
-import {vbbProfile} from 'hafas-client/p/vbb.js'
+import {profile} from 'hafas-client/p/vbb.js'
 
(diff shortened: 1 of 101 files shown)


derhuerst:
Thanks for pointing this out!

Can you adapt the `import`s to still keep the profile's name in the variable? I think it makes the code slightly more understandable. E.g. `import {profile as dbProfile} from 'hafas-client/p/db/index.js'`. 🙏

KristjanESPERANTO:
Ok, I adjusted it like this :slightly_smiling_face:

KristjanESPERANTO:
I just stumbled across a typo, sorry. But now I should be done.

derhuerst:
Thank you! I have squashed and merged your commits in d8805d9.

KristjanESPERANTO:
Nice. Should I do an extra PR for the typo? I obviously committed it too late.

derhuerst:
I have cherry-picked the typo fix as 1fdb7882.

PR #280, opened 2022-11-22
Update actions version

KristjanESPERANTO:
With this we get rid of a few warning messages during the tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,9 +10,9 @@ jobs:
         node-version: [16.x, 18.x]
     steps:
     - name: checkout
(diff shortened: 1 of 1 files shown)


derhuerst:
> With this we get rid of a few warning messages during the tests.

Do you have an example for such warnings?

KristjanESPERANTO:
[In the annotations of an other PR](https://github.com/public-transport/hafas-client/actions/runs/3526730580) you can see the deprecation warnings.

[In the annotations of this PR](https://github.com/public-transport/hafas-client/actions/runs/3526703109) are no warnings.

PR #292, opened 2023-05-11
apply leg-wide remarks if `opt.stopovers` is false

traines-source:
I noticed that when stopovers=false, the remarks are missing entirely from journey legs.

Was there any reason to not just parse them? Maybe because it gets too bloated (loads of Komfort-Checkins, bicycle conveyance etc.)?


derhuerst:
> I noticed that when `stopovers=false`, the remarks are missing entirely from journey legs. Was there any reason to not just parse them? […]

I can't remember any reason. Thanks for this PR.

test/db-journey-polyline.js
```diff
@@ -23,7 +23,7 @@ const opt = {
 	bike: false,
 	tickets: true,
 	polylines: true,
-	remarks: true,
+	remarks: false,
```
derhuerst, line 26: I assume that you change `opt.remarks` to `false` whenever adding the expected remarks to the fixture(s) would have been too cumbersome?

traines-source: Yes, at least it was easier to set it to `false`, and I figured it would not be a bad idea to have some variance and also test the `false` behaviour somewhere

traines-source:
Btw as far as I could tell the integration tests failed already before for some other reason, I didn't do anything about that

derhuerst:
Do you want to be mentioned as a [contributor](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#people-fields-author-contributors)? If you do, with which name and/or email?

Issue #287, opened 2023-04-02
DB: make routing mode configurable, so that cancelled trips can be queried & pagination works


lamBOOO:
I noticed that some Deutsche Bahn journeys are missing in the response. Especially, canceled trains seem to not show up.

For example, the ICE 595 from Mannheim to Stuttgart today at 12:30 (`2023-04-02T12:30:00+02:00`) is missing in the response, although it is listed in the `reiseauskunft.bahn.de` and in the DB Navigator App (see images below).

I try to use the API with the following code:
```JS
let {createDbHafas} = await import('db-hafas');  // REPL
let hafas = createDbHafas('test@test.com')
let journeys = await hafas.journeys('8000244', '8000096', {
	results: 2,
	departure: new Date("2023-04-02T12:14:00+02:00")
})
console.log(JSON.stringify(journeys, null, 4))
```

The result is:
```JSON
{
    "laterRef": "2|OF|MT#14#164953#164953#164995#164995#0#0#5#164895#2#0#1050#0#0#-2147483648#1#2|PDH#07b4c623b5c8a2cb5cd37e278d01c93a|RD#2042023|RT#121000|US#0",
    "journeys": [
        {
            "type": "journey",
            "legs": [
                {
                    "origin": {
                        "type": "stop",
                        "id": "8000244",
                        "name": "Mannheim Hbf",
                        "location": {
                            "type": "location",
                            "id": "8000244",
                            "latitude": 49.479181,
                            "longitude": 8.469268
                        },
                        "products": {
                            "nationalExpress": true,
                            "national": true,
                            "regionalExpress": true,
                            "regional": true,
                            "suburban": true,
                            "bus": true,
                            "ferry": false,
                            "subway": false,
                            "tram": true,
                            "taxi": false
                        }
                    },
                    "destination": {
                        "type": "stop",
                        "id": "8000096",
                        "name": "Stuttgart Hbf",
                        "location": {
                            "type": "location",
                            "id": "8000096",
                            "latitude": 48.785052,
                            "longitude": 9.182589
                        },
                        "products": {
                            "nationalExpress": true,
                            "national": true,
                            "regionalExpress": true,
                            "regional": true,
                            "suburban": true,
                            "bus": true,
                            "ferry": false,
                            "subway": false,
                            "tram": true,
                            "taxi": false
                        }
                    },
                    "departure": "2023-04-02T12:14:00+02:00",
                    "plannedDeparture": "2023-04-02T12:14:00+02:00",
[...]

PR #295, opened 2023-07-18
Add routingModes option to db profile

lamBOOO:
Sorry for the delay, but I started working on #287. It seems to work so far.

I'm a little confused with the testing, so I "only" added an e2e to test all different routingMode options for errors.

Is the testing done properly?

### DIFF
--- a/lib/db-routing-modes.js
+++ b/lib/db-routing-modes.js
@@ -1,11 +1,11 @@
 // see https://pastebin.com/qZ9WS3Cx
 const routingModes = {
 	OFF: 'OFF',
-  INFOS: 'INFOS',
-  FULL: 'FULL',
-  REALTIME: 'REALTIME',
-  SERVER_DEFAULT: 'SERVER_DEFAULT',
-  HYBRID: 'HYBRID',
+	INFOS: 'INFOS',
+	FULL: 'FULL',
+	REALTIME: 'REALTIME',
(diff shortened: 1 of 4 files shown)


derhuerst:
Thanks!

I have squashed your commits into one, and adapted that one commit slightly:

- moved `lib/db-routing-modes.js` to `p/db/routing-modes` (if it's DB-specific, let's put it in `p/db`; if it's not, let's rename it and put it in `lib`)
- added a integration test fixture (which is just a set of recorded E2E test HTTP requests)
- minor code formatting

lamBOOO:
Thank you. 👍

Just for understanding: The integration test fixture is generated by "simply" running the `test-integration:record` script and then pushing the updated `recording.har` into the repo? 🤔

derhuerst:
> The integration test fixture is generated by "simply" running the `test-integration:record` script and then pushing the updated `recording.har` into the repo? 🤔

Yes, although – as you have pointed out – running `test-integration:record` will fail in practice because most test suites/files' `T_MOCK` are in the past, so querying new HTTP responses from HAFAS won't work.

I have documented some of the gotchas of the testing setup in 19cdde06. If you have more questions (or even want to contribute more guidance), don't hesitate to create a new Issue!

lamBOOO:
Thanks, it's now more clear to me. My initial starting point for testing/contribution was the `contributing.md`.

great work btw 👍

PR #305, opened 2023-12-03
Update ESLint and add Stylistic Formatting

KristjanESPERANTO:
~~Prettier and ESLint are a nice combination.~~

**Update**: Because prettier hardly allows any customization, we switched to stylistic in the process of the review.


KristjanESPERANTO:
The tests are failing because I didn't run `npm run lint:fix` intentionally.

derhuerst:
A appreciate the intention to make the code's formatting consistent and reproducible, but infortunately in the past years, whenever I have encountered prettier-formatted code, prettier's formatting style was really incompatible with the line-based nature of Git and trying to minimise unrelated changes in commits.

I'm aware that I'm letting technical limitations of Git dictate my code style here, but given that I'm bound to use Git, the benefits of having Git-diff-friendly code formatting are significant: Less conflicts while rebasing/merging, more meaningful `git blame` to remember why a change was made, etc.

The problem that I have with automated formatting: It depends on a specific code section's *business logic* how likely it is to be extended in the future (e.g. an another argument added to a function call, or another field to an object), and the current tools cannot estimate that. For example a `addNumbers(a, b)` call is unlikely to be extended in the future, a `sendRequest(url, body, headers)` much more.

If we accept a "dumb" and therefore inherently Git-diff-unfriendly code formatting tool, let's pick one – or a well-supported way to get prettier to do this – that does the following:

- always use trailing commas in arrays, arguments lists, etc.
- long arithmetic expressions, ternary expressions, arguments lists, arrays of non-primitives, objects, destructuring assignments, `export` expressions, etc: almost always on multiple lines instead of one, to make future changes have nicer Git diffs
- always use curly braces for multi-line blocks (`if`/`else`, `while`, etc)
- when >1 function calls are chained, never inline them, but always give each call a separate line
- don't inline multi-line empty objects if they are nested in a data structure (e.g. [here](https://github.com/public-transport/hafas-client/blob/9a3a1d3b0ce9b36bbce23ae251df5760f5b67e36/lib/errors.js#L71-L72))

derhuerst (changes requested):
The linting setup currently produces output like this:

```js
const locHints = (l.remarkRefs || []).
	filter(ref => !!ref.hint && Array.isArray(ref.tagL)).
	filter(({tagL}) => tagL.includes('RES_LOC') ||
	tagL.find(t => t.slice(
		0,
		8,
	) === 'RES_LOC_'), // e.g. `RES_LOC_H3`
	// eslint-disable-next-line @stylistic/function-paren-newline
	).
	map(ref => ref.hint)
```

Following the goals I outlined before, I would like it to look as follows:

```js
const locHints = (l.remarkRefs || [])
	.filter(ref => !!ref.hint && Array.isArray(ref.tagL))
	.filter(({tagL}) => (
		tagL.includes('RES_LOC') ||
		tagL.find(t => t.slice(0, 8) === 'RES_LOC_' // e.g. `RES_LOC_H3`
	))
	.map(ref => ref.hint)
```

(I wouldn't really mind the first `.filter()` call being broken up into multiple lines; Same for the `.find()` call. But trust people (contributors and me) to decide case-by-case, and would therefore turn off the rule about this.)

---

Sorry to put the tedious task of configuring a linter on you…

index.js
```diff
@@ -75,19 +77,32 @@ const createClient = (profile, userAgent, opt = {}) => {
 			stopovers: false, // fetch & parse previous/next stopovers?
 			// departures at related stations
 			// e.g. those that belong together on the metro map.
-			includeRelatedStations: true
-		}, opt)
+			includeRelatedStations: true,
+		},
+		opt)
```
derhuerst, line 82: I dislike the style, but for the sake of consistency, can we get all of thee  arguments indented and with a trailing comma?

index.js
```diff
@@ -75,19 +77,32 @@ const createClient = (profile, userAgent, opt = {}) => {
...
 
-		const req = profile.formatStationBoardReq({profile, opt}, station, type)
-
-		const {res, common} = await profile.request({profile, opt}, userAgent, req)
-
-		const ctx = {profile, opt, common, res}
-		const jnyL = Array.isArray(res.jnyL) ? res.jnyL : []
-		const results = jnyL.map(res => parse(ctx, res))
-		.sort((a, b) => new Date(a.when) - new Date(b.when)) // todo
+		const req = profile.formatStationBoardReq({profile,
```
derhuerst, line 86: Sorry, this is just weird. 😅

index.js
```diff
@@ -75,19 +77,32 @@ const createClient = (profile, userAgent, opt = {}) => {
...
+
+		const ctx = {profile,
+			opt,
+			common,
+			res}
+		const jnyL = Array.isArray(res.jnyL)
+			? res.jnyL
+			: []
+		const results = jnyL.map(res => parse(ctx,
+			res)).
```
derhuerst, line 104: As reviewed before: Leading `.`, one method call per line.

derhuerst (changes requested):

KristjanESPERANTO:
I see it as a learning process, don't worry.

Sorry, I just realized that I have to go back a few steps. Please wait with the review.

KristjanESPERANTO:
I think I'm a bit closer to this what you want, but now I have to figure out why the tests are failing here. Locally they are not failing. I'll try to continue in the next few days.

index.js
KristjanESPERANTO: Done by rule `"@stylistic/dot-location": ["error", "property"]`.

index.js
KristjanESPERANTO: Undone with last changes.

index.js
KristjanESPERANTO: Now it looks like that:

```js
			includeRelatedStations: true,
		}, opt)
```

Do you want a trailing comma after opt?

KristjanESPERANTO:
Okay, the unit tests are now running :rocket:  and I think we are now close to your specifications. I'm delighted that stylistic is so customizable :smiley:  There's certainly still potential for some fine-tuning.

Would you like to take a look? There are changes to only 171 files 😅

p/bvg/index.js
```diff
@@ -19,8 +19,8 @@ import {products} from './products.js'
 // todo: there's also a referenced icon `{"res":"occup_fig_{low,mid}"}`
 const addOccupancy = (item, occupancyCodes) => {
 	const remIdx = (item.remarks || [])
-	.findIndex(r => r.code && occupancyCodes.has(r.code))
-	if (remIdx < 0) return;
+		.findIndex(r => r.code && occupancyCodes.has(r.code))
+	if (remIdx < 0) return
```
derhuerst, line 23: An aspect that I forgot to mention: I like to put `;` behind `return` so that, if the next line starts with an expression (e.g. `foo.bar()`), I don't accidentally return the result of the expressions, but `void`.

Is it hard to add this rule?

p/db/ageGroup.js
```diff
@@ -9,22 +9,17 @@ const ageGroup = {
...
-		return ageGroup.BABY
-	if (age < upperBoundOf.CHILD)
-		return ageGroup.CHILD
-	if (age < upperBoundOf.YOUNG)
-		return ageGroup.YOUNG
-	if (age < upperBoundOf.ADULT)
-		return ageGroup.ADULT
-	if (age < upperBoundOf.SENIOR)
-		return ageGroup.SENIOR
+	if (age < upperBoundOf.BABY) return ageGroup.BABY
```
derhuerst, line 18: I guess we can just always use `{`/`}` for conditionals & loop blocks.

p/bvg/index.js
KristjanESPERANTO: The options for setting semicolons are "always" (default) and "never" (the current setting). I don't understand overriding rules well enough to say whether it is possible to create an exception for result with acceptable effort.
We could also disable the rule, but then it becomes inconsistent again.

We could also switch off the rule completely.

lib/request.js
```diff
@@ -57,7 +57,8 @@ const randomizeUserAgent = (userAgent) => {
 	return ua
 }
 
-const md5 = input => createHash('md5').update(input).digest()
+const md5 = input => createHash('md5').update(input)
```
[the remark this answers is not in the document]
KristjanESPERANTO: I handled this with commit 37e9e48. I don't know if we want this behavior in all places. At least it would be consistent :slightly_smiling_face:

p/db/ageGroup.js
KristjanESPERANTO: I handled this with commit 300a3e4.

format/date.js
```diff
@@ -4,16 +4,18 @@ import {luxonIANAZonesByProfile as timezones} from '../lib/luxon-timezones.js'
...
 	}
 
-	return DateTime.fromMillis(+when, {
+	return DateTime.fromMillis(Number(when), {
 		locale: profile.locale,
-		zone: timezone
-	}).toFormat('yyyyMMdd')
+		zone: timezone,
+	})
+		.toFormat('yyyyMMdd')
```
derhuerst, line 18: Is reasonably easy to either a) indent *all* chained function calls or b) *none*?

format/location-filter.js
```diff
@@ -1,6 +1,14 @@
 const formatLocationFilter = (stops, addresses, poi) => {
-	if (stops && addresses && poi) return 'ALL'
-	return (stops ? 'S' : '') + (addresses ? 'A' : '') + (poi ? 'P' : '')
+	if (stops && addresses && poi) {
+		return 'ALL'
+	}
+	return (stops
```
derhuerst, line 5: I guess this is not easy to make more readable while not breaking other section's formatting? 😄

derhuerst (commented):
I'm happy with almost all formatting! Just some small items left.

format/date.js
KristjanESPERANTO: That rule caused that rule: <https://eslint.style/rules/js/newline-per-chained-call>

a) As I understand it, it is currently set as this. The new line is created for all chains. Even if the chain only has one element. The default is from 2 elements.

b) We could switch off the rule, so that all variants would be possible, but this could make the code inconsistent.
Setting the rule so that all chains, no matter how long they are, are in one line (what I understand by b), I don't think works. And I know no other suitable rule for this.

format/location-filter.js
KristjanESPERANTO: We could switch of the curly rule: <https://eslint.org/docs/latest/rules/curly>. I think it would be perfect fine if we don't force curly braces.

Then it could remain a one-liner, if that's what you mean by "more readable"(?) :slightly_smiling_face:

derhuerst: No, I mean the (to me) unintuitive multi-line formatting of the ternary expressions.

For example, indent the "components" of a ternary expression (condition, truthy expression, falsy expression) intended, each on a separate line, and the wrapping parantheses one indentation less.

format/date.js
derhuerst: > a) As I understand it, it is currently set as this. The new line is created for all chains. Even if the chain only has one element. The default is from 2 elements.

Let's try enabling it for >=1 elements then.

p/bvg/index.js
derhuerst: I don't mind either way, pick one. I'm used to omitting semicolons, as they're *slightly* less visual noise, but they really help prevent bugs as explained above, so it's a tie for me.

derhuerst:
Because I couldn't find a better way to contact you, so I'll ask here, independently of this PR: What do you think about becoming a co- oder fallback maintainer (up to you which) in case I end up not being able to take care of `hafas-client` anymore? If you have more questions or comments on this, don't hesitate to [contact me](https://jannisr.de/), e.g. via email!

p/bvg/index.js
KristjanESPERANTO: Okay, I see the point of visual noise, but prevent bugs is more significant. I also like to use the defaults and in this case semicolons are set by default. I just made a commit that adjusts the rule. Of course, this affects a huge amount of lines in the code.

format/date.js
KristjanESPERANTO: Okay, the rule is set now this way:

`"@stylistic/newline-per-chained-call": ["error", { "ignoreChainWithDepth": 1 }],`

format/location-filter.js
KristjanESPERANTO: You're talking about this part, right?

```js
	return (stops
		? 'S'
		: '') + (addresses
		? 'A'
		: '') + (poi
		? 'P'
		: '');
};
```

You are right the formatting was not very readable. I couldn't get a better version without a bit manual work. These two variants would be possible:

1. The new current state with this rule `"@stylistic/multiline-ternary": ["error", "always-multiline"]`:

```js
	return (
		(stops ? 'S' : '')
		+ (addresses ? 'A' : '')
		+ (poi ? 'P' : '')
	);
```

2. This would be possible with this rule `"@stylistic/multiline-ternary": ["error", "always"]`:

```js
	return (
		(stops
			? 'S'
			: '')
			+ (addresses
				? 'A'
				: '')
				+ (poi
					? 'P'
					: '')
	);
```

format/date.js
derhuerst: The `.fromMillis()` in `DateTime.fromMillis().toFormat()` is not indented. My idea was to put *every* call in a chain on its own line as soon as there are >1 calls in the chain.
I don't mind wether it's indented or not.

KristjanESPERANTO: Okay, I didn't expect that. But I took another look at the description of the rule. The following examples are listed as OK:

```js
// Or
_
  .chain({})
  .map(foo)
  .filter(bar);

// Or
_.chain({})
  .map(foo)
  .filter(bar);
```

I have not yet found a suitable setting or rule for automatically forcing the first linebreak.

I have now adjusted it manually, but that doesn't quite fit in with what we want to achieve with the use of a linter.

format/location-filter.js
KristjanESPERANTO: Yes, definitely `always-multiline` :+1:

format/date.js
derhuerst: Still, it's an improvement over the un-linted status quo!

derhuerst (approved):

KristjanESPERANTO:
:partying_face:

PR #308, opened 2023-12-20
Fix mode for 'ec-ic'

KristjanESPERANTO:
This will fix on error in the integration test.

The tap update was only suitable here, it has nothing to do with the fix.
Update: The tap update has been withdrawn.

### DIFF
--- a/p/mobil-nrw/products.js
+++ b/p/mobil-nrw/products.js
@@ -67,7 +67,7 @@ const products = [
 	},
 	{
 		id: 'ec-ic',
-		mode: 'ec-ic',
+		mode: 'train',
 		bitmasks: [2],
 		name: 'EC/IC',
 		short: 'EC/IC',


derhuerst:
> The tap update was only suitable here, it has nothing to do with the fix.

The change from `^18.0.4` to `^18.6.1` doesn't have any effects since a) we're using a semver range here (`^`) and b) the lockfile isn't checked into Git.

If you had cloned `hafas-client` and installed dependencies (which creates a `package-lock.json` lockfile) before, and now want to get the latest `tap` version, run `npm upgrade` instead.

p/mobil-nrw/products.js
```diff
@@ -67,7 +67,7 @@ const products = [
 	},
 	{
 		id: 'ec-ic',
-		mode: 'ec-ic',
+		mode: 'train',
```
derhuerst, line 70: Even though `ec-ic` was not intended, this could be considered a breaking change by some people who depend on this specific value. I'm still fine with publishing with a non-major version though.

KristjanESPERANTO:
Okay I've withdrawn the tap update.

derhuerst:
Thanks!

derhuerst:
In case anyone is wondering: A while ago, I published this fix as `hafas-client@6.2.1`. 🎉

PR #313, opened 2024-01-26
fixed incorrect parameters in documentation

Revilum:
I'm not sure how exactly the `query` parameter works exactly, but the source code for `client.lines` requires the parameter `query` and checks if the string is empty.

### DIFF
--- a/docs/lines.md
+++ b/docs/lines.md
@@ -1,4 +1,4 @@
-# `lines([opt])`
+# `lines(query, [opt])`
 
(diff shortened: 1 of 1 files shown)


derhuerst:
Thanks!

Unfortunately, AFAIK it is HAFAS-endpoint-specific what can be passed into `query`, so I haven't properly documented it. If you want, you're very welcome to document some examples with some specific profile(s).

derhuerst:
@Revilum Do you still want to add some examples here? Otherwise I'll merge this as-is.

Revilum:
I have only tried this with the nah.sh endpoint (which requires it) but yes you can merge this if you want.

PR #314, opened 2024-03-20
DB: support "best price search"

### LINKED ISSUE #291: DB: support "best price search"
Both the *Deutsche Bahn* (DB) website as well as their app have a feature called "best price search" or "Bestpreissuche", which works just like the regular `journeys()` (`TripSearch` in HAFAS), except that it returns the lowest-price journeys for each time bracket.

The details of this `BestPriceSearch` call are documented at https://github.com/juliuste/db-prices/issues/33#issuecomment-1538776446 . It's response data looks like the one of a `TripSearch` call, but it also has a field `res.outDaySegL` that groups the journeys in `res.outConL[]` into time brackets.

Not sure how this functionality should be exposed:
- We could let `journeys()` always parse & expose a grouping/bracketing of journeys as soon as the HAFAS response contains it. A DB-specific `opt.bestPriceSearch` option would use `BestPriceSearch` and `outTime: "000000"`.
- Within the DB profile, we could implement it as an entirely separate `hafas-client` method. It would have to share much of the logic with `journeys()`. Currently, we don't have a mechanism to let profiles add methods though.

---

pyhafas equivalent Issue: https://github.com/FahrplanDatenGarten/pyhafas/issues/26

derhuerst:
@bergmannjg I went ahead and created a WIP PR.

Fixes #291.


index.js
```diff
@@ -245,7 +245,79 @@ const createClient = (profile, userAgent, opt = {}) => {
...
+			jnyFltrL: filters,
+			getTariff: !!opt.tickets,
+
+			getPolyline: !!opt.polylines
+		}
+		query.outDate = profile.formatDate(profile, when)
+
+		const {res, common} = await profile.request({profile, opt}, userAgent, {
+			cfg: {polyEnc: 'GPA'},
+			meth: 'BestPriceSearch',
```
derhuerst, line 302: I assume this call is only supported by the DB HAFAS endpoint?
If that is the case, we should make it `db`-profile-only (need to think about how).
If that is not the case, let's keep it in here, but also check *which* endpoints actually support it.

bergmannjg: I checked with cli.js that BestPriceSearch is only valid for the db profile.

> If that is the case, we should make it db-profile-only

Maybe it is sufficient to add a test that the current profile is the db profile.

It is ok that bestPrices is part of the optional api for hafasclient like trip or radar?

parse/bestprice.js
```diff
@@ -0,0 +1,26 @@
...
+			.map(i => journeys.find(j => j.refreshToken == res.outConL[i].ctxRecon))
+			.filter(j => Boolean(j))
+		: [];
+
+	const amount = outDaySeg.bestPrice.amount / 100;
+	const currency = bpjourneys?.[0]?.price?.currency;
+
+	const result = {
+		journeys: bpjourneys,
+		fromDate: profile.parseDateTime(ctx, outDaySeg.fromDate, outDaySeg.fromTime),
```
derhuerst, line 16: Given that these two fields also contain a time (that is not always `00:00`, according to the test fixture), I propose renaming them to either `{from,to}DateTime` or just `{from,to}`. What do you think?

index.js
```diff
@@ -268,6 +269,88 @@ const createClient = (profile, userAgent, opt = {}) => {
 		};
 	};
 
+	const bestPrices = async (from, to, opt = {}) => {
+		from = profile.formatLocation(profile, from, 'from');
+		to = profile.formatLocation(profile, to, 'to');
+
+		opt = Object.assign({
+			via: null, // let journeys pass this station?
+			transfers: -1, // maximum nr of transfers
```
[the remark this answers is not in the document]
bergmannjg: > Are there more options that BestPriceSearch supports that are not listed here?

I think the options for BestPriceSearch are via, transfers, transferTime, bike, products and trfReq. Added missing option transferTime.

index.js
```diff
@@ -268,6 +270,87 @@ const createClient = (profile, userAgent, opt = {}) => {
...
+		}
+
+		let when = new Date();
+		if (opt.departure !== undefined && opt.departure !== null) {
+			when = new Date(opt.departure);
+			if (Number.isNaN(Number(when))) {
+				throw new TypeError('opt.departure is invalid');
+			}
+			const now = new Date();
+			let today = DateTime.fromObject({
```
derhuerst, line 300: Sorry, even though I told you to do it this way, I *think* this isn't correct either: For example, if `profile.timezone` is `Europe/Berlin`, the timezone-naive `now` is `2024-05-06T00:30+01:00` because it runs on a machine in `Europe/London`, then this will create a Berlin-timezone timestamp for the next day.

I can think of a least two options, the latter being the more straightforward one IMO:

1. Change `now` to be timezone-aware, so that `now.{year, month,day}` are correct.
2. Just use one chain of `DateTime` calls, as illustrated below.

Also, while implicitly casting `today` to a `number` below works, it would be cleaner to add `.toMillis()`.

```js
const today = DateTime
	.fromMillis(Date.now(), {
		zone: profile.timezone,
		locale: profile.locale,
	})
	.startOf('day')
	.toMillis()
```

index.js
```diff
@@ -268,6 +269,88 @@ const createClient = (profile, userAgent, opt = {}) => {
...
+		}
+
+		let when = new Date();
+		if (opt.departure !== undefined && opt.departure !== null) {
+			when = new Date(opt.departure);
+			if (Number.isNaN(Number(when))) {
+				throw new TypeError('opt.departure is invalid');
+			}
+			const now = new Date();
+			const today = new Date(now.getFullYear(), now.getMonth(), now.getDate());
```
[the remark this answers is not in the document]
derhuerst: see https://github.com/public-transport/hafas-client/pull/314#discussion_r1590392396

PR #332, opened 2024-12-08
Upgrade ESLint

KristjanESPERANTO:
1. Upgrade ESLint to v9
2. Switch from deprecated eslintrc config to flat config
3. Apply rules to codebase
4. The linter workflow didn't work after the ESLint upgrade, so I did two things to fix this:
    1. Move the linter test to a separate job that runs only once. It is not necessary to run it with every node version.
    2. Since the GitHub action [complained about the missing lock file](https://github.com/public-transport/hafas-client/actions/runs/12225229939/job/34098967427#step:3:25), I added it to the repository. According to [the node documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json), it is recommended to commit the lock file anyway: _"This file is intended to be committed into source repositories, and serves various purposes..."_.

Should I rather outsource the adjustments to the workflows to an extra PR?


.github/workflows/test.yml
```diff
@@ -6,6 +6,22 @@ env:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v4
+      - name: "Use Node.js"
+        uses: actions/setup-node@v4
+        with:
+          node-version: 23
+          cache: "npm"
```
derhuerst, line 19: Apparently this causes the "Dependencies lock file is not found" error.Can you remove it, as well as the lockfile?

Given that `hafas-client` is a library (and not an app/service), the lockfile would only apply when cloning it and installing dependencies via `npm install`, so it has very limited benefits. At the same time, it is cumbersome to work with it. This is why I'd like the repo not to have a lockfile.

.github/workflows/test.yml
```diff
@@ -6,6 +6,22 @@ env:
...
+  lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v4
+      - name: "Use Node.js"
+        uses: actions/setup-node@v4
+        with:
+          node-version: 23
```
derhuerst, line 18: Please use an LTS version.

```suggestion
          node-version: 'lts/*'
```

.github/workflows/test.yml
```diff
@@ -6,6 +6,22 @@ env:
...
+          cache: "npm"
+      - name: "Install dependencies"
+        run: |
+          npm install
+      - name: "Run linter tests"
```
derhuerst, line 23:
```suggestion
      - name: "Run linter"
```

derhuerst (changes requested):

.github/workflows/test.yml
KristjanESPERANTO: > Apparently this causes the "Dependencies lock file is not found" error.

Ah makes sense :bulb: 

> Can you remove it, as well as the lockfile?

Yes and done :slightly_smiling_face:

derhuerst (approved):
This looks quite good.

Do you know how much work it would be to get rid of the remaining stylistic changes? I don't want you to put *even more* work into wrangling `@stylistic/*`, but maybe there are some low-hanging fruit?

.github/workflows/test.yml
KristjanESPERANTO: Nice. I didn't know this option :rocket:

.github/workflows/test.yml
KristjanESPERANTO: Done.

KristjanESPERANTO:
I think we have adapted the rules quite well to this project and your reasonable preferences.

But I understand the desire to define as few exceptions as possible.

I see three options:

1. Leave it as it is for the time being.
2. Instead of activating all `@stylistic` rules, we could only activate the recommended ones. With this we would have to define far fewer exceptions.
3. Another option would be to use ‘[@antfu/eslint-config](https://github.com/antfu/eslint-config)’ instead of all other ESLint plugins. The package is from a/the `@stylistic` maintainer and comes with a huge number of opinionated rules. With this the ESLint configuration file would only consist of two lines (if we don't define exception). I'm currently considering converting some projects to this. I see two main advantages: A very simple ESLint configuration and a lot of rules (not only for JavaScript files). If one does not like the settings of several rules, this of course would also bloat the config.

I would be open to all three (or you see another another one). But I would prefer to complete this PR so we can check the upgrade to ESLint 9 :slightly_smiling_face: 

Then I could create pull requests for option 2 and 3 with manageable effort to see what the changes would look like if we don't define exceptions.

KristjanESPERANTO:
> Do you know how much work it would be to get rid of the remaining stylistic changes?

Oh, I got you wrong! But now I got it. Yes, I think by disabling or adapting one or two rules, we can get rid of these stylistic changes. I will soon see into it.

p/rsag/example.js
```diff
@@ -11,7 +11,7 @@ const albertEinsteinStr = {
 	type: 'location',
 	id: '990004201',
 	address: 'Rostock - Südstadt, Albert-Einstein-Straße 23',
-	latitude: 54.077208, longitude: 12.108299
+	latitude: 54.077208, longitude: 12.108299,
```
KristjanESPERANTO, line 14: Actually this seems to be a good change. For example if someone will later add a line, the diff will be clearer. Do you also want to get rid of changes like this?

derhuerst: It's indeed in line with what I explained in https://github.com/public-transport/hafas-client/pull/305#issuecomment-1837485331, let's keep it.

derhuerst (changes requested):
I think this is ready to merge, except that `package-lock.json` is still checked in (I'd like to keep it out).

KristjanESPERANTO:
> I think this is ready to merge, except that `package-lock.json` is still checked in (I'd like to keep it out).

Just removed it. I reset my branch for that and used this chance to fix typos in my commit messages 😅

derhuerst (approved):
Thanks! I have slightly renamed your commits.

PR #339, opened 2025-04-02
fix: added esm js files for module imports in environments without createRequire

crwntec:


crwntec:
### Update
I have replace the imports for all profile with ESM syntax and removed the unused require

test/db-deps-with-destination.js
```diff
@@ -29,7 +29,15 @@ tap.test('parses departure.destination correctly (DB)', (t) => {
...
 	t.ok(departure.destination, 'missing departure.destination');
-	t.equal(departure.destination.type, 'stop', 'invalid departure.destination.type');
-	t.equal(departure.destination.id, '930200', 'invalid departure.destination.id');
+	t.equal(
+		departure.destination.type,
+		'stop',
+		'invalid departure.destination.type',
+	);
+	t.equal(
+		departure.destination.id,
```
derhuerst, line 38: Please don't commit all the unrelated styling changes, this makes reviewing and [later `git blame`-ing](https://engineering.leanix.net/blog/essence-of-git-commit-message/) much harder.

crwntec: Sorry VSCode somhow auto applied formatting when replacing and pasting. Was a bit of a pain to rollback and revert those changes on my fork but it passes all tests again and imports the profiles correctly

crwntec:
I fixed the accidental styling/formatting changes and voila down from 3000+ changed lines to just 850

crwntec (commented):
All tests passed

crwntec:
I think so but to be safe I left them in there for now. Also not all require statements have been ported to imports yet. Would this also be necessary? Because the error message only refers to `base.json` imports

derhuerst:
> I think so but to be safe I left them in there for now. Also not all require statements have been ported to imports yet. Would this also be necessary? Because the error message only refers to `base.json` imports

Whenever you change a file from `.json` to `.js`, all `require()` calls to it should be replaced with a import statements.

crwntec:
Ok so no other Import statements required for now. What about the other json files. Do you have any plans on switching completely to import statements?

derhuerst:
I think we can switch over to ESM files, instead of keeping both.

crwntec:
New PR for this? It's a bit more work but I could look into porting everything else

derhuerst:
No, I think you're misunderstanding. Please just remove all `p/*/base.json` files in favour of the new ESM ones.

p/avv/index.js
```diff
@@ -1,9 +1,9 @@
 // todo: use import assertions once they're supported by Node.js & ESLint
 // https://github.com/tc39/proposal-import-assertions
-import {createRequire} from 'module';
-const require = createRequire(import.meta.url);
+// import {createRequire} from 'module';
```
derhuerst, line 3: These should be removed, too.

crwntec: All commented out createRequires?

crwntec:
Everything correct now?

p/avv/index.js
```diff
@@ -1,9 +1,8 @@
 // todo: use import assertions once they're supported by Node.js & ESLint
 // https://github.com/tc39/proposal-import-assertions
```
KristjanESPERANTO, line 2: It looks good to me! :clap: But can't these two lines also be removed now? :mag:

crwntec: Not really imo because import assertions are not used throughout the project. They seem to be somewhat experimental and my proposed changes are using a js file exporting JSON so no assertion is needed. Assertions are required when using an import statment on a JSON file.

KristjanESPERANTO:
The index.js in the vbb profile still uses the json file.

crwntec:
Somhow one of the newer commits didn't contain all changes so certain profiles weren't changed. Should be working now

p/avv/index.js
crwntec: However it is only the tests that are importing with require() now so this is something that could be changed in the future

derhuerst: > Not really imo because import assertions are not used throughout the project.

Please remove this comment. I'd like to keep only in places where `require()` is actually used.

> They seem to be somewhat experimental [...[.

[They'll likely be part of ECMAScript 2025](https://github.com/tc39/proposals/blob/be6d67902372fa9006537f3178736a88a46ad375/finished-proposals.md?plain=1#L12). But even then, `hafas-client`'s minimum Node.js version must support them.

derhuerst (approved):
I have squashed your commits into one, and amended it with a few changes,
- reverting changes to unrelated files, e.g. stylistic changes
- restoring some `hafas-client`-specific changes on top of what `pull-profile-base-data.sh` pulls from [`transport-apis`](https://github.com/public-transport/transport-apis) that were there before. (I now, not very contribution-friendly!)
- removing some leftover comments and empty lines, in order to keep the set of changed lines strictly related to the JSON -> ESM switch.

derhuerst:
Fix published as `hafas-client@6.3.5`. 🎉

crwntec:
Ok nice. Massive thank you for the support. I will test out the integration with my [application](https://db-live-monitor.vercel.app/)!

PR #347, opened 2025-12-30
fix(test): rescue BVG radar test with near-future timestamp

KristjanESPERANTO:
BVG's radar API returns empty movements for timestamps too far in the future. The test used createWhen() which calculates "next Monday 10am", causing the API to return no vehicle data.

Solution: Use +1 hour timestamp for E2E tests (live API), keep standard `when` for integration tests (matches existing fixtures):

```js
    const radarWhen = process.env.VCR_MODE && !process.env.VCR_OFF
      ? when
      : new Date(Date.now() + 60 * 60 * 1000);
```

### DIFF
--- a/test/e2e/bvg.js
+++ b/test/e2e/bvg.js
@@ -384,16 +384,28 @@ tap.test('stop', async (t) => {
 });
 
 tap.test('radar', async (t) => {
+	// BVG's radar API doesn't return movements for timestamps too far in the
+	// future. For E2E tests, use a near-future timestamp (1 hour). For integration
+	// tests, use the standard `when` to match existing fixtures.
+	const radarWhen = process.env.VCR_MODE && !process.env.VCR_OFF
+		? when
+		: new Date(Date.now() + 60 * 60 * 1000);
+	const radarValidators = createVbbBvgValidators({when: radarWhen});
+	const validateRadar = createValidate(radarValidators.cfg, {
+		movement: radarValidators.validateMovement,
+	});
+
 	const res = await client.radar({
 		north: 52.52411,
 		west: 13.41002,
 		south: 52.51942,
 		east: 13.41709,
 	}, {
-		duration: 5 * 60, when,
+		duration: 5 * 60,
+		when: radarWhen,
 	});
 
-	validate(t, res, 'radarResult', 'res');
+	validateRadar(t, res, 'radarResult', 'res');
 	t.end();
 });
 


test/e2e/bvg.js
```diff
@@ -384,16 +384,20 @@ tap.test('stop', async (t) => {
...
 tap.test('radar', async (t) => {
+	// BVG's radar API doesn't accept timestamps too far in the past,
+	// so we don't pass `when` and just validate the structure
 	const res = await client.radar({
 		north: 52.52411,
 		west: 13.41002,
 		south: 52.51942,
 		east: 13.41709,
 	}, {
-		duration: 5 * 60, when,
```
derhuerst, line 393: Unfortunately, this change [breaks the integration tests](https://github.com/public-transport/hafas-client/actions/runs/20606497620/job/59183256721?pr=347#step:6:361).

The integration tests are the E2E tests, except that they are run against a prerecorded set of HTTP responses, and the HTTP requests need to (understandably) match exactly, so a deterministic/static `when` needs to be included. This is why `createWhen()` uses `T_MOCK` if `VCR_MODE` is defined, and just uses `Date.now()` otherwise.

derhuerst (changes requested):

test/e2e/bvg.js
derhuerst: TLDR: With the E2E tests (`npm run test-e2e`), this test should *already* run with `Date.now()`. And because the integration tests (`npm run test-integration`) run against previously-recorded HTTP response fixtures, it doesn't matter that the BVG API wouldn't be able to handle the hard-coded `T_MOCK` value (because it is too far in the past).

KristjanESPERANTO: Oh, why I didn't catch that the integration tests failed? :facepalm: 

The issue is that even though `createWhen()` correctly uses `Date.now()` in E2E mode, it calculates "next Monday 10am" from that, which is a **future** timestamp. BVG's radar API doesn't return movements for future dates.

I've made the test mode-aware: integration tests use `when` with full validation (fixtures), E2E tests omit `when` to get current vehicle data.

Updated commit and PR description. What do you think?

derhuerst: The current implementation does not test if `radar()`'s `opt.when` parameter works. Instead, we could pass a `when` that is less far in the future (e.g. in 1 hour).

However, I'm fine with merging this, it is good enough IMHO.

KristjanESPERANTO: Done! Used +1h for E2E tests as suggested :smiley:

derhuerst: Thanks!

derhuerst (approved):

